PR #33, opened 2024-05-14
:construction_worker: Add initial CI 

evaline-ju:


gkumbhat (approved):
Looks good to me!

gkumbhat:
@evaline-ju please do test it in `main` and see if the build passes. We can open up another PR if it doesn't work and iterate.

Issue #34, opened 2024-05-15
Add CI


evaline-ju:
## Description

As an orchestrator developer I want tests and ideally formatting/linting checks to run on each build automatically so developers can ensure that any changes pass existing tests/checks and reviewers have more confidence in changes.

## Acceptance Criteria

- [ ] Github workflow to run tests and formatting

PR #36, opened 2024-05-15
:construction_worker::bug: Use docker for test run

evaline-ju:
Formatting, linting and tests all require protoc and compilation of files. Since the goal of this repo is to output an image, and we are already installing protoc in the Docker image, we can essentially reuse parts of the image build. A failure at any stage will lead to "build" failure.

Usage of GHA caching taken from https://github.com/IBM/text-generation-inference. Docker build actions are used in order to standardize docker usage and allow for caching. The file was renamed since the focus is on "building" to allow for testing and formatting, etc., but a future build workflow could `build` from the release tag and potentially push an image to a registry

For: #34


gkumbhat (approved):
LGTM

Issue #16, opened 2024-05-03
Host API in swagger pages in the repo


gkumbhat:
### Description

To enable easy readability of the API, we want to host the swagger UI in github pages for the orchestrator API. We should also try if there is a way to host multiple APIs, that way we can also host detectors API in this repo as well. But first priority is to host orchestrator API.

PR #38, opened 2024-05-20
Add swagger UI

gkumbhat:
closes: #16 

### Changes

- Add swagger UI to host and publish orchestrator and detector APIs

There is a lot of code that I got from `swagger-ui` following [this](https://github.com/peter-evans/swagger-github-pages) tutorial. But files I changed are:
1. `index.html`
2. `dist/swagger-initializer.js`
3. `docs/api/orchestrator_openapi_0_1_0.yaml`
4. `docs/api/openapi_detector_api.yaml`

The pages are available for demo here: https://gkumbhat.github.io/fms-guardrails-orchestrator/?urls.primaryName=Orchestrator+API


gkumbhat:
@evaline-ju yep. yaml changes should automatically get reflected. If we need to change some swagger ui elements, like UI Name, attribute name or add another API spec, then we would just add them to list in `dist/swagger-initializer.js` and it would automatically get built, deployed and available in pages.

Issue #35, opened 2024-05-15
Add user request validation


evaline-ju:
## Description

As an orchestrator user, I want to receive errors on invalid requests, so that I can know how to potentially update my request. 

## Discussion

Scenarios include:
- Unsupported LLMs (model IDs) (this may be handled by #13 / PR #31)
- Unsupported detectors (this may be handled by #13 / PR #31)
- Incorrect text generation parameters (this may be handled by #13 / PR #31) [500s should be 400s]
- Missing parameters e.g. users can put in `"input": { "model-name": {} }` without the `models` parameter and receive no error. The `model-name` would then just be ignored on input, which seems unexpected.
- Extra parameters e.g. users can put in random additional fields to the overall request or under `guardrail_config`. These extra fields are currently ignored but could indicate unintentional misformatting.
- Should `masks` without `models` on input detection even be valid?

## Acceptance Criteria

- [ ] Handling of above cases
- [ ] Unit tests

PR #42, opened 2024-05-22
:safety_vest: Initial request validation

evaline-ju:
- Update API documentation to reflect updated error handling. Auto-generated Fast API objects no longer needed
- Provide some upfront request validation of masks and required parameters fields, with some unit tests 
- Update types - fields that cannot be less than 0 are now updated to be unsigned. `models` are required (removed `Option`) under `input` and `output` and will now result in a validation error if not present

Part of: #35 

[Please squash merge!]

### DIFF
--- a/src/orchestrator.rs
+++ b/src/orchestrator.rs
@@ -478,9 +478,9 @@ async fn generate(
             Ok(ClassifiedGeneratedTextResult {
                 generated_text: Some(response.text.clone()),
                 finish_reason: Some(response.stop_reason().into()),
-                generated_token_count: Some(response.generated_token_count as u32),
+                generated_token_count: Some(response.generated_token_count),
                 seed: Some(response.seed as u32),
-                input_token_count: response.input_token_count as u32,
+                input_token_count: response.input_token_count,
                 warnings: None,
                 tokens: if response.tokens.is_empty() {
                     None


declark1 (commented):
Hmm, I'm personally not a fan of adding garde for validation unless it provides some feature(s) that we really need. Rust's type system can be leaned on for many things, such as:

```rust
#[garde(range(min = 0))]
pub max_new_tokens: Option<i32>,
```
Rather, this should just be an unsigned `Option<u32>` instead of checking `min = 0`.

Below, a type is being marked as "required" despite it being an `Option` field, so the intent isn't clear. We should just leverage Rust's `Option` when things are optional, all else being required.

```rust
 #[serde(rename = "models")]
    #[garde(required)] // output field must have `models`
    pub models: Option<HashMap<String, DetectorParams>>,
```
Also, the garde derive macros add (even more) clutter to the struct fields (especially `#[garde(skip)` and the mysterious `#[garde(dive)]` which I see added to many fields).

For stuff like below, I'd say implementing simple rules in the validate method is better.

instead of this:
```rust
   #[garde(length(min = 1))]
    pub model_id: String,
```

we can do this:
```rust
pub fn validate(&self) -> Result<(), Error> {
   if self.model_id.is_empty() {
      return Error::Validation("`model_id` must be provided")
   }
   // other rules
}
```

The "newtype" pattern is also common in idiomatic Rust for cases like the non-empty `String` (if reused a lot), e.g. you may create a `NonEmptyString` newtype unit struct that wraps a `String` to provide the non-empty guarantee. 
- Of course there are crates for the common ones like this, e.g. https://docs.rs/non-empty-string/latest/non_empty_string/struct.NonEmptyString.html, but it's easy enough to implement too.

As a side-effect of deriving `garde::Validate`, I see you had to name the validate method `upfront_validate()` and call into `self.validate()` to run garde's validation. IMO, it feels kind of blackbox'ish vs. a chain of explicit rules.

declark1:
> > Below, a type is being marked as "required" despite it being an `Option` field, so the intent isn't clear. We should just leverage Rust's `Option` when things are optional, all else being required.
> 
> Sure, part of the thinking here had been if some of these API structs were to be updated potentially via auto-generation when the API yamls were updated, the structs themselves with `Option` could kept, while the requirements of the orchestrator API itself could be enforced through validation. We could certainly encode some of the rules into the types themselves (unsigned, no `Option` etc.)
> 
> I agree that the verboseness of the `skip` needed on fields to not validate was not ideal in the case where we had fairly large structs. `[garde(dive)]` was meant to provide validation for nested/inner structs.

Yea, the auto-generation is also bad as it's generating a lot of garbage such as unnecessary field renames (the fields already have these names), unnecessary `Option`'s, unnecessary derives, signed integers for things that should be unsigned, etc. models.rs` could be a lot cleaner. Considering these types don't appear hard to maintain at all, i.e. they don't have 100 fields, do we really need to auto-generate going forward? If not, we could clean these up.

declark1:
Just thought of something. Won’t auto-generation overwrite all of these changes to the model types, such as the garde derives, validation method, etc?

declark1 (approved):
This looks good to me. I think this makes sense to keep it simple instead of using a validation crate.

I typically avoid using errors from other modules (`server::Error`) unless its a crate-level error shared by all modules, but it isn't a big deal here as the `models` module can really be considered part of the `server`.

Another option is to crate a `ValidationError` type in `models.rs`:

```rust
#[derive(Debug, thiserror::Error)]
pub enum ValidationError {
    #[error("`{0}` is required")]
    Required(String),
    #[error("{0}")]
    Invalid(String),
    // etc
}
//...
```
and then in `server.rs`, impl a conversion into `server::Error` so that you can still call `request.validate()?`:

```rust
//...
impl From<models::ValidationError> for Error {
    fn from(value: models::ValidationError) -> Self {
        Self::Validation(value.to_string())
    }
}
```

Issue #41, opened 2024-05-22
Empty masks array leads to different behavior/results on unary endpoint


evaline-ju:
## Describe the bug

The presence of the an empty `masks` array should lead to the same behavior as if the `masks` array is left off completely. Both indicate that there are no spans that should be specifically split out of the `inputs` text.

## Sample Code

Replace `detector_name` and `model_name` as appropriate for the inputs, and port as appropriate for the local server

```
curl -v -H "Content-Type: application/json" --request POST \
--data '{"model_id": "model_name",
    "inputs": "There was once a really dumb cat. The chicken crossed the road. The social security number is 123-45-6789.",
    "guardrail_config": {
        "input": {
            "models": {
                "detector_name": {}
            }, "masks": []
        },
        "output": {"models":{}}
    },
    "text_gen_parameters": {
        "min_new_tokens": 40,
        "max_new_tokens": 200
    }
    }' \
    http://localhost:<port>/api/v1/task/classification-with-text-generation
```
vs.
```
curl -v -H "Content-Type: application/json" --request POST \
--data '{"model_id": "model_name",
    "inputs": "There was once a really dumb cat. The chicken crossed the road. The social security number is 123-45-6789.",
    "guardrail_config": {
        "input": {
            "models": {
                "detector_name": {}
            }
        },
        "output": {"models":{}}
    },
    "text_gen_parameters": {
        "min_new_tokens": 40,
        "max_new_tokens": 200
    }
    }' \
    http://localhost:<port>/api/v1/task/classification-with-text-generation
```

## Expected behavior

Same results

## Observed behavior

Different results [potentially a some vs. none case]

Example output with empty `masks` array in input
```
[...]

PR #43, opened 2024-05-22
:bug: Empty masks same as no masks

evaline-ju:
Tested with calls in #41 - now return the same response

Closes: #41


declark1 (approved):
LGTM!

PR #49, opened 2024-05-24
:technologist: Replace content endpoint with batch endpoint

gkumbhat:
### Changes
- Replace `content` endpoint to support `batch` operation by default.


gkumbhat:
I think L13 is still accurate since this endpoint is designed for detectors that work on user prompt or generated text. Its just that instead of 1 document, now the API accepts multiple documents.

Issue #24, opened 2024-05-09
Deal with the Threshold filter from detectors at Orchestrator


alanbraz:
based on discussions around user requirements, detectors should not deal with logic of filtering results based on user's threshold, it should be done at the higher level

PR #52, opened 2024-05-29
:sparkles: Add thresholding for detector results

evaline-ju:
Following #51 , score thresholding for detector results will be done by the orchestrator. Each detector will need a configured `default_threshold`, expected to be added when the detector is added to the orchestrator config. 

If a user provides a threshold through the API, this will be used. Otherwise the default threshold will be used.

Example config parts API that should give the same output if `default_threshold`: 0.5 were configured for detector `model_name`
```
"guardrail_config": {
        "input": { "models": {}},
        "output": {"models":{"model_name": {}}}
    },
```
```
"guardrail_config": {
        "input": { "models": {}},
        "output": {"models":{"model_name": {"threshold": 0.5}}}
    },
```

This PR also adds documentation/comments for the config structs and updates the validation error for request validation from #42. 

For: #24

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -137,13 +137,6 @@ impl OrchestratorConfig {
             .get(detector_id)
             .map(|detector_config| detector_config.chunker_id.clone())
     }
-
-    /// Get default threshold of a particular detector
-    pub fn get_default_threshold(&self, detector_id: &str) -> Option<f32> {
-        self.detectors
-            .get(detector_id)
-            .map(|detector_config| detector_config.default_threshold)
-    }
 }
 
 fn service_tls_name_to_config(
--- a/src/orchestrator.rs
+++ b/src/orchestrator.rs
@@ -238,21 +238,19 @@ async fn detect(
             let ctx = ctx.clone();
             let detector_id = detector_id.clone();
             let detector_params = detector_params.clone();
+            // Get the detector config
+            let detector_config =
+                ctx.config
+                    .detectors
+                    .get(&detector_id)
+                    .ok_or_else(|| Error::DetectorNotFound {
+                        detector_id: detector_id.clone(),
+                    })?;
             // Get the default threshold to use if threshold is not provided by the user
-            let default_threshold =
-                ctx.config
-                    .get_default_threshold(&detector_id)
-                    .ok_or_else(|| Error::DetectorNotFound {
-                        detector_id: detector_id.clone(),
-                    })?;
+            let default_threshold = detector_config.default_threshold;
             // Get chunker for detector
-            let chunker_id =
-                ctx.config
-                    .get_chunker_id(&detector_id)
-                    .ok_or_else(|| Error::DetectorNotFound {
-                        detector_id: detector_id.clone(),
-                    })?;
-            let chunks = chunks.get(&chunker_id).unwrap().clone();
+            let chunker_id = detector_config.chunker_id.as_str();
+            let chunks = chunks.get(chunker_id).unwrap().clone();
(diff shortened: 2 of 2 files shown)


src/config.rs
```diff
@@ -6,70 +6,100 @@ use std::{
...
+/// Configuration for each detector
 #[derive(Debug, Clone, Deserialize)]
 pub struct DetectorConfig {
+    /// Detector service connection information
     pub service: ServiceConfig,
+    /// ID of chunker that this detector will use
     pub chunker_id: String,
-    //pub config: HashMap<String, String>,
+    /// Optional detector configuration parameters
+    pub config: Option<DetectorConfigParams>,
```
declark1, line 84: Can we drop `config: Option<DetectorConfigParams>` and add `default_threshold` to `DetectorConfig`? The nested config object seems unnecessary.

I also think we should make it `f32` instead of `Option<f32>` since its a default.

src/orchestrator.rs
```diff
@@ -237,7 +237,13 @@ async fn detect(
...
             let ctx = ctx.clone();
             let detector_id = detector_id.clone();
-            let detector_params = detector_params.clone();
+            let mut detector_params = detector_params.clone();
+            if let Some(default_threshold) = ctx.config.get_default_threshold(&detector_id) {
+                // Use a default threshold if threshold is not provided by the user
+                detector_params
+                    .entry("threshold".into())
+                    .or_insert(default_threshold.into());
+            }
```
declark1, line 246: Since the orchestrator handles applying the threshold anyway, I would just add a `default_threshold` parameter to `handle_detection_task()` instead of inserting it into the `DetectorParams` map, unless there are plans for that to change.

Then, in the results `filter_map` you can do...

If we change `default_threshold` to `f32`:
```rust
  //...
  let threshold = detector_params
      .get("threshold")
      .and_then(|v| v.as_f64())
      .unwrap_or(default_threshold as f64);
  (result.score >= threshold).then_some(result)
```

Or if we keep it as `Option<f32>`:
```rust
  //...
  let threshold = detector_params
      .get("threshold")
      .and_then(|v| v.as_f64())
      .or(default_threshold.map(Into::into));
  if threshold.is_some_and(|value| result.score < value) {
      None
  } else {
      Some(result)
  }
```
(we can also get rid of the casts if we change it to `f64`)

src/orchestrator.rs
```diff
@@ -238,6 +238,14 @@ async fn detect(
             let ctx = ctx.clone();
             let detector_id = detector_id.clone();
             let detector_params = detector_params.clone();
+            // Get the default threshold to use if threshold is not provided by the user
```
declark1, line 241: Now that we need two items from the detector config, I would just `get` the detector config once (returning an error if it's not found) and get `default_threshold` and `chunker_id` from it. One less lookup and then you don't have to do `.ok_or_else(|| Error::DetectorNotFound {}))` twice.

You can also drop that `OrchestratorConfig::get_default_threshold()` method then.

e.g.
```rust
// Get the detector config
let detector_config =
    ctx.config
        .detectors
        .get(&detector_id)
        .ok_or_else(|| Error::DetectorNotFound {
            detector_id: detector_id.clone(),
        })?;
// Get the default threshold to use if threshold is not provided by the user
let default_threshold = detector_config.default_threshold;
// Get chunker for detector
let chunker_id = detector_config.chunker_id.as_str();
let chunks = chunks.get(chunker_id).unwrap().clone();
```

declark1 (commented):
Looks good, just once last suggested change inline.

declark1 (approved):
LGTM!

PR #61, opened 2024-06-04
:bug: Evidences as array

evaline-ju:
`evidences` by its plurality was meant to be an array, but this was unclear from the spec


gkumbhat (approved):
Looks good to me!

Issue #37, opened 2024-05-16
Update Detectors API


alanbraz:
## Description

As a developer, I want to call any detector's API, so that I can use it correctly in my workflow

## Discussion

API updated and will need to refactor orchestrator code to accommodate them.

## Acceptance Criteria

- [ ] update openapi yaml
- [ ] update code to call detector that implements the new API

PR #62, opened 2024-06-04
:sparkles: Update to `/text/contents` API for detector

evaline-ju:
Updates the detector client for the newly decided `/text/contents` endpoint at https://foundation-model-stack.github.io/fms-guardrails-orchestrator/?urls.primaryName=Detector+API

One of the main differences is the detector would no longer return the text corresponding to spans, i.e. what is used to fill in `word` in the final API response today

Optimizing `DetectorParams` formatting has been left out for now here in order to get the functionality updates

Closes: #37

### DIFF
--- a/src/clients/detector.rs
+++ b/src/clients/detector.rs
@@ -28,7 +28,7 @@ impl DetectorClient {
             .clone())
     }
 
-    pub async fn analyze_contents(
+    pub async fn text_contents(
         &self,
         model_id: &str,
         request: ContentAnalysisRequest,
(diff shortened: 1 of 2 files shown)


src/clients/detector.rs
```diff
@@ -47,39 +47,71 @@ impl DetectorClient {
...
-
-impl From<Detection> for crate::models::TokenClassificationResult {
-    fn from(value: Detection) -> Self {
+impl From<ContentAnalysisResponse> for crate::models::TokenClassificationResult {
+    fn from(value: ContentAnalysisResponse) -> Self {
         Self {
             start: value.start as u32,
             end: value.end as u32,
-            word: value.text,
+            word: "".to_string(), // Expected to be filled in
```
gkumbhat, line 114: nit: Can you add a TODO here that this value will be provided by the detectors API in next iteration so we won't have to do this for long?

gkumbhat (commented):
small nit. But otherwise LGTM

declark1 (approved):
Pushed a few changes for efficiency and to avoid extra allocations. Renamed a few things.

Issue #56, opened 2024-05-31
Document streaming result aggregation policy in ADR


evaline-ju:
## Description

As an orchestrator architect, I want to document the rationale for how streamed detector results will be aggregated while keeping track of span ordering.

## Discussion

Provide detailed discussion here

## Acceptance Criteria

- [x] ADR exists under https://github.com/foundation-model-stack/fms-guardrails-orchestrator/tree/main/docs/architecture/adrs

PR #63, opened 2024-06-04
:memo: Document streaming response strategy in ADR

evaline-ju:
Closes: #56


docs/architecture/adrs/002-streaming-response-aggregation.md
```diff
@@ -0,0 +1,43 @@
...
+    3. To ensure there are no premature responses e.g. maximum `processed_index` is 42 and another queue has responses at `processed_index` 5, 17, and 34, responses will be “awaited” for that detector until a response with `processed_index` reaches 42 or more. If more than 42, the last response will be left on the queue.
+    4. Once all the responses for up to a certain maximum `processed_index` are identified, the detections will be aggregated into a list to be returned in the `token_classification_results` field of the frame to be streamed back. The text of that frame will of course correspond to the text of the frame that had the maximum `processed_index`.
+    5. Once a frame is returned/streamed back, the process will repeat to look for at least one queued response from each detector for comparison, until each detector has processed to the end of the stream.
+4. The aggregation strategy can be configurable, even if only one is available and implemented to begin with.
```
gkumbhat, line 31: We should write it as an implementation / design of the streaming logic inside orchestrator. Since currently it looks like these would be configurable by the user via API.

docs/architecture/adrs/002-streaming-response-aggregation.md
```diff
@@ -0,0 +1,43 @@
+# ADR 002: Streaming Response Behavior and Result Aggregation
+
+This ADR defines the orchestrator strategy of result aggregation and behavior observed by end users of the initial streaming API endpoint (i.e. streaming text generation with classification) when detections are made on streamed output of text generation models.
+
```
gkumbhat, line 4: Can we make a note that these are only considering the `/contents` detectors and not other ATM

docs/architecture/adrs/002-streaming-response-aggregation.md
```diff
@@ -0,0 +1,43 @@
...
+## Motivation
+
+This orchestrator is designed to work with various detectors. Each detector will have an associated chunker indicated in an end user configuration. Each chunker could chunk different amounts of text e.g. a sentence chunker vs. a paragraph chunker. Each detector could then provide detection responses on different amounts of text. In the unary case, all of these detection responses could be aggregated together and returned in the single response to the end user. 
+
+However, in the streaming case, as frames are being streamed back, the end user will need to receive detection responses corresponding to the text in the streamed frames. Frames still have to be returned in order so that the user can parse the stream of generated text and detections and know how much of the original text generation stream has been processed i.e. what part of the stream that detections have been run on.
+
+
+## Decisions
+
+### End user experience
```
gkumbhat, line 14: Can we add a note on how we will receive and send back `input_token_count` ?

docs/architecture/adrs/002-streaming-response-aggregation.md
```diff
@@ -0,0 +1,43 @@
...
+
+1. Only fully processed frames will be returned to the end user. This means that for each (streamed) frame returned to the user, all detector responses for that particular frame's text up to that frame’s index (`processed_index`) will be returned with the frame.
+    1. Generated texts in output frames with detections are still expected to be contiguous, meaning the end user can still piece all of the frames together. This aligns with the way various text generation servers provide output.
+2. In order to achieve (1) for detectors that may work with different chunkers, e.g. paragraph vs. sentence, results will not be streamed back/returned until the “slowest” detector has processed. “Slowest” may not necessarily just be in terms of time but in terms of amount of text processed (i.e. largest `processed_index`).
```
gkumbhat, line 18: One of the implication here is also that the "default document chunker" (or no chunker) that is within orchestrator, will also have to provide response in the datastructure of streaming 🤔

docs/architecture/adrs/002-streaming-response-aggregation.md
```diff
@@ -0,0 +1,43 @@
...
+2. In order to achieve (1) for detectors that may work with different chunkers, e.g. paragraph vs. sentence, results will not be streamed back/returned until the “slowest” detector has processed. “Slowest” may not necessarily just be in terms of time but in terms of amount of text processed (i.e. largest `processed_index`).  
+
+### Response aggregation strategy
+
+1. Only chunker calls will provide indices of the text generation server frames processed (`start_index` and `processed_index`). Detector calls are expected to be unary, and responses only indicate the detections on any given chunk from a chunker. Detectors are expected to still give a reply (i.e. empty array) if there are no detections on a given chunk.
+    1. On any detector response, the orchestrator will have to associate the chunker indices with each detection response. The orchestrator will update any span start/end on detection responses with chunker indices to be able to indicate to the end user where the detection was in the overall stream.
+2. The orchestrator will have to maintain detection responses that are not ready to be streamed back to the user, such as with queues.
```
gkumbhat, line 24: Can we make a note here about how this will work along with thresholding? Will we store results after applying threshold in the queue or before applying threshold?

gkumbhat (commented):
Thanks for adding the ADR Evaline. Added few comments and questions.

docs/architecture/adrs/002-streaming-response-aggregation.md
gkumbhat: Yeah, I think thresholding them first make sense.  But in that case, would we not put it in queue if the response turns out to be not relevant ? Since in that case, how would be know that the detector has responded, because then there are 2 scenarios that can cause the response to be not in the queue, i.e detector has not responded and detector responded but the response was not relevant. In both cases, we would end up waiting ?

gkumbhat: sounds good! Thanks

gkumbhat (approved):
LGTM

Issue #64, opened 2024-06-06
Additional validation on threshold parameter


evaline-ju:
## Description

As a orchestrator developer, I want to validate the `threshold` field, so that I can prevent users from putting in non-sensical values for threshold, like strings. This includes potentially using a specific struct for detector parameters.

## Discussion

Currently the detector parameters are very flexible serde JSON values: https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/cdbb37c972351a3728f528f47029f22d5b62dd37/src/models.rs#L6-L10 

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

PR #66, opened 2024-06-06
:goal_net::label: Use more defined struct for threshold

evaline-ju:
Detector parameters provided for each detector will not be passed to each detector client currently. The one expected parameter currently is `threshold`, used by the orchestrator to potentially filter each detector's results. Including this named parameter allows for a more defined type and validation (as opposed to the `serde_json` value that would have allowed comparison of scores to strings)

Closes: #64

### DIFF
--- a/TODOs.md
+++ b/TODOs.md
@@ -1,14 +1,8 @@
 # TODOs
-- [ ] Design detector map config and implement loading this config file
 - [ ] Add TLS support for orchestrator server
 - [ ] Implement Health probes
-- [ ] Add request models (objects for request and response)
 - [ ] Add unit tests
-- [ ] Add request validation for classification with text generation endpoint
 - [ ] Add request validation for streaming classification with text generation endpoint
-- [ ] Host API in swagger pages in the repo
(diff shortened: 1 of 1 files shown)


declark1 (commented):
Should we just make threshold `f64` everywhere to avoid the casts to `f64`?

src/models.rs
```diff
@@ -3,11 +3,12 @@
...
-// TODO: When detector API is updated, consider if fields
-// like 'threshold' can be named options instead of the
-// use a generic HashMap with Values here
-// ref. https://github.com/foundation-model-stack/fms-guardrails-orchestrator/issues/37
-pub type DetectorParams = HashMap<String, serde_json::Value>;
+/// Parameters relevant to each detector
+#[derive(Clone, Debug, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct DetectorParams {
+    /// Threshold with which to filter detector results by score
+    pub threshold: Option<f32>,
```
gkumbhat, line 10:
```suggestion
    pub threshold: Option<f64>,
```

declark1 (approved):

PR #67, opened 2024-06-07
Renamed index_codepoints to slice_codepoints and updated to do it cleaner

declark1:
I thought of a more efficient way to do this. Renamed to `slice_codepoints`. Also updated `apply_masks` to use `slice_codepoints`.


gkumbhat (approved):
This looks way more efficient. 
`skip` and `take` both are `O(1)` right ?

PR #76, opened 2024-06-12
:technologist: Add text to contents analysis detectors response API

gkumbhat:
### Changes
- Add `text` in detector API


alanbraz (approved):
looks good

Issue #74, opened 2024-06-11
 v2.0 finish_reason responds with EOS_TOKEN instead of MAX_TOKENS.


flaviabeo:
## Describe the bug

When running functional tests, calling the `/api/v1/task/classification-with-text-generation` end point through:

```python
response = requests.post(url="https://gr2-fmaas-tuning.apps.fmaas-devstage-backend.fmaas.res.ibm.com/api/v1/task/classification-with-text-generation", json=PAYLOAD, headers=HEADERS, verify=SSL)
json_response = json.loads(response.content.decode(response.encoding))
```

## Platform

Please provide details about the environment you are using, including the following:

- fms-orchestrator image version: `us.icr.io/cil15-shared-registry/fms-orchestr8:latest`

## Expected behavior

The previous response at v1.0 was:
```json
{
  "generated_text": "I love dogs love@dogs.com",
  "token_classification_results": {
    "input": null,
    "output": []
  },
  "finish_reason": "MAX_TOKENS",
  "generated_token_count": 99,
  "seed": 42,
  "input_token_count": 21,
  "warnings": null,
  "tokens": [
    {
      "text": "\u0120not",
      "logprob": 0.0,
      "rank": 1
    }],
  "input_tokens": []
  }
```
`finish_reason` value is expected to be "MAX_TOKENS".

## Observed behavior

The actual response is:
```json
{
  "generated_text": "I hate cats they are stupid. I love dogs love@dogs.com",
  "token_classification_results": {},
  "finish_reason": "EOS_TOKEN",
  "generated_token_count": 59,
  "seed": 0,
  "input_token_count": 21
}
```
`finish_reason` value is "EOS_TOKEN".

## Additional context

The functional test run to get this error is:

```python
    def test_unary_no_guardrails():
        response = requests.post(url=UNARY_GUARDRAILS_URL, json=PAYLOAD, headers=HEADERS, verify=SSL)
        assert 200 == response.status_code
        json_response = json.loads(response.content.decode(response.encoding))
>       assert json_response["finish_reason"] == "MAX_TOKENS"
E       AssertionError: assert 'EOS_TOKEN' == 'MAX_TOKENS'
E         - MAX_TOKENS
E         + EOS_TOKEN
```

declark1:
[...]

PR #80, opened 2024-06-13
Fix decode_method mapping

mdevino:
This PR addresses issue #74.

### DIFF
--- a/src/models.rs
+++ b/src/models.rs
@@ -583,7 +583,10 @@ impl From<ExponentialDecayLengthPenalty> for pb::fmaas::decoding_parameters::Len
 impl From<GuardrailsTextGenerationParameters> for pb::fmaas::Parameters {
     // NOTE: types should really be consistent between APIs
     fn from(value: GuardrailsTextGenerationParameters) -> Self {
-        let decoding_method = value.decoding_method.unwrap_or("GREEDY".to_string());
+        let mut decoding_method = value.decoding_method.unwrap_or("GREEDY".to_string());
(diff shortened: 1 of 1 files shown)


src/models.rs
```diff
@@ -583,7 +583,10 @@ impl From<ExponentialDecayLengthPenalty> for pb::fmaas::decoding_parameters::Len
 impl From<GuardrailsTextGenerationParameters> for pb::fmaas::Parameters {
     // NOTE: types should really be consistent between APIs
     fn from(value: GuardrailsTextGenerationParameters) -> Self {
-        let decoding_method = value.decoding_method.unwrap_or("GREEDY".to_string());
+        let mut decoding_method = value.decoding_method.unwrap_or("GREEDY".to_string());
+        if decoding_method == "SAMPLING" {
+            decoding_method = String::from("SAMPLE");
+        }
```
mdevino, line 589: I am yet to create the unit test for this.
Is there a _rustier_ way of addressing this, though?

gkumbhat: hmm. actually why is there `SAMPLING` vs `SAMPLE` discrepancy. We should just use the name used by TGIS 🤔

mdevino: This seems to be the only place in the code that uses it. Let's close this issue and make sure we call the API using "SAMPLE" instead of "SAMPLING".

mdevino:
After some discussion, this code is no longer needed. Let's close this issue and make sure we call the API using "SAMPLE" instead of "SAMPLING".

Issue #14, opened 2024-05-02
Add TLS support for orchestrator [server-side]


gkumbhat:
### Description

We need to add support for the orchestrator server to allow serving the endpoints over mutual TLS.  

There are a couple TODOs left in the code, such as unused parameters for running the server. With the client path CA info, mTLS will be enabled.

### Acceptance criteria
- [ ] Orchestrator server works with tls configurations and accepts requests only over https
- [ ] Tls configuration tested 
- [ ] How to configure server with TLS config documented in README

PR #82, opened 2024-06-14
:lock: Add server-side mTLS

evaline-ju:
## Details
- Moved `/health` endpoint to separate endpoint where TLS won't apply so that this can potentially be used by k8s probes for server healthchecks
- Add TLS/mTLS ability to boot a https server, keeping graceful shutdown - if a client CA is provided, this will configure mTLS. If only cert, key are provided, this will configure TLS. If none are provided, this will boot the insecure http server. 
- There were challenges using `axum-server`, which used an older `rustls` version at time of implementation, leading to conflicts on the `ServerConfig` object that would hold TLS info. Thus the decision was made to use lower-level hyper/tower (that axum is built on) to implement the TLS capabilities. 

## Testing
With self-signed certs via curl (so `-k` used)

Setting orchestrator server-side certs
```
export TLS_CLIENT_CA_CERT_PATH=/path/to/ca.crt
export TLS_KEY_PATH=/path/to/localhost.key
export TLS_CERT_PATH=/path/to/cert.pem
```
With properly signed certs
```
curl --cert client_0.crt --key client_0.key --cacert ca.crt -k -v -H "Content-Type: application/json" --request POST .... https://localhost:8081/api/v1/task/classification-with-text-generation
```
Response:
```
> POST /api/v1/task/classification-with-text-generation HTTP/1.1
> Host: localhost:8081
> User-Agent: curl/7.84.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 447
>
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* old SSL session ID is stale, removing
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* old SSL session ID is stale, removing
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* old SSL session ID is stale, removing
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< content-type: application/json
< content-length: 772
< date: Fri, 14 Jun 2024 00:23:44 GMT
<
* Connection #0 to host localhost left intact
{"token_classification_results": ...
```

With improperly signed certs with another CA
```
curl --cert client_02.crt --key client_02.key --cacert ca.crt -k -v -H "Content-Type: application/json" --request POST ...     https://localhost:8081/api/v1/task/classification-with-text-generation
```
Response
```
> POST /api/v1/task/classification-with-text-generation HTTP/1.1
> Host: localhost:8081
> User-Agent: curl/7.84.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 447
>
* TLSv1.3 (IN), TLS alert, unknown CA (560):
* OpenSSL SSL_read: error:14094418:SSL routines:ssl3_read_bytes:tlsv1 alert unknown ca, errno 0
* Closing connection 0
curl: (56) OpenSSL SSL_read: error:14094418:SSL routines:ssl3_read_bytes:tlsv1 alert unknown ca, errno 0
```

Closes: #14

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -128,7 +128,7 @@ pub async fn run(
     let listener: TcpListener = TcpListener::bind(&http_addr)
         .await
         .unwrap_or_else(|_| panic!("failed to bind to {http_addr}"));
-    let handle = if arc_server_config.is_some() {
+    let guardrails_handle = if arc_server_config.is_some() {
         // TLS
         // Use more low level server configuration than axum for configurability
         // Ref. https://github.com/tokio-rs/axum/blob/main/examples/low-level-rustls/src/main.rs
@@ -213,9 +213,9 @@ pub async fn run(
     };
 
     // (3) Launch each server as a separate task
-    let (health_res, res) = tokio::join!(health_handle, handle);
+    let (health_res, guardrails_res) = tokio::join!(health_handle, guardrails_handle);
     health_res.unwrap();
-    res.unwrap();
+    guardrails_res.unwrap();
     info!("Shutdown complete for servers");
     Ok(())
 }


src/server.rs
```diff
@@ -33,20 +41,64 @@ pub struct ServerState {
...
+    tls_key_path: Option<PathBuf>,
+    tls_client_ca_cert_path: Option<PathBuf>,
     config_path: PathBuf,
 ) -> Result<(), Error> {
     let config = OrchestratorConfig::load(config_path).await;
     let orchestrator = Orchestrator::new(config).await?;
     let shared_state = Arc::new(ServerState { orchestrator });
 
-    // TODO: configure server TLS
+    // Separate HTTP health server without TLS for probes
```
gkumbhat, line 54: Can you divide all the different functionalities in this function into logic steps and document them here. So sort of like:
```
1. Configure handlers with routes
2. Configure TLS if requested with each of the routes
...
```
and then in your comments just above those implementations, you can also associated same step number. That way it would be a bit more easy to follow

src/server.rs
```diff
@@ -60,15 +124,99 @@ pub async fn run(
...
         .with_state(shared_state);
 
-    let listener = TcpListener::bind(&http_addr)
+    // (2c) Generate main guardrails server handle based on whether TLS is needed
+    let listener: TcpListener = TcpListener::bind(&http_addr)
         .await
         .unwrap_or_else(|_| panic!("failed to bind to {http_addr}"));
-    let server =
-        axum::serve(listener, app.into_make_service()).with_graceful_shutdown(shutdown_signal());
+    let handle = if arc_server_config.is_some() {
```
gkumbhat, line 131: nit: can we change this to `app_handle` or something?

gkumbhat (approved):

gkumbhat (approved):
Left small nit, but otherwise, LGTM

PR #92, opened 2024-06-20
:bug: Provide a default crypto provider

evaline-ju:
Crypto provider apparently needs a default

Ref: 
- https://github.com/rustls/hyper-rustls/blob/main/examples/server.rs
- https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html


gkumbhat (approved):
Looks good to me!

Issue #44, opened 2024-05-22
Implement server side streaming for text generation 


gkumbhat:
## Description

As a user, I want to handle server side streaming for text generation, instead of unary request, so that I can utilize orchestration in more interactive use-cases.

This story is to wire up text generation streaming with guardrails orchestrator streaming endpoint. Connecting it with chunker and detection will be handled in separate issue

PR #95, opened 2024-06-21
:construction::sparkles: Initial streaming text generation through streaming endpoint

evaline-ju:
Adds basic streaming text generation to the streaming classification endpoint

Note: detections are not enabled yet, and `start_index` and `processed_index` will have to be updated

For: #44


gkumbhat (approved):
We need to iterate on a few things w.r.t to channel etc, but this is a good start and would allow us to work on a couple of pieces in parallel. 

Thanks Evaline!

Issue #96, opened 2024-06-25
Detected PII word's "start" and "end" are returning the wrong positions


flaviabeo:
## Describe the bug

The start and end fields returned are different from expected. For example, the e-mail is not in the mask position returned by the request response from the detected PII.

## Platform

Please provide details about the environment you are using, including the following:

GR Version 2.0 NLP Client, TLS
 
## Sample Code

POST call to `/api/v1/task/classification-with-text-generation` with the payload:

```json
{
    "inputs": "I hate cats they are stupid. I love dogs love@dogs.com. Rabbits are pretty but",
    "model_id": "bloom-560m",
    "guardrail_config": {
        "input": {
            "models": {
                "en_syntax_rbr_pii": {
                    "threshold": 0.8
                }
            }
        },
        "output": {
            "models": {}
        }
    },
    "text_gen_parameters": {
        "preserve_input_text": true,
        "max_new_tokens": 99,
        "min_new_tokens": 1,
        "truncate_input_tokens": 500,
        "decoding_method": "SAMPLING",
        "top_k": 2,
        "top_p": 0.9,
        "typical_p": 0.5,
        "temperature": 0.8,
        "seed": 42,
        "repetition_penalty": 2,
        "max_time": 0,
        "stop_sequences": [
            "42"
        ]
    }
}
```

## Expected behavior

```json
{
        "start": 41,
        "end": 54,
        "word": "love@dogs.com",
        "entity": "EmailAddress",
        "entity_group": "",
        "score": 0.8
  }
```

## Observed behavior

```json
{
    "token_classification_results": {
        "input": [
            {
                "start": 12,
                "end": 25,
                "word": "they are stup",
                "entity": "EmailAddress",
                "entity_group": "pii",
                "score": 0.8
            }
        ]
    },
    "input_token_count": 21,
    "warnings": [
        {
            "id": "UNSUITABLE_INPUT",
            "message": "Unsuitable input detected. Please check the detected entities on your input and try again with the unsuitable input removed."
        }
    ]
}
```

gkumbhat:
Thanks @flaviabeo . This should have been fixed with: https://github.com/foundation-model-stack/fms-guardrails-orchestrator/pull/47

flaviabeo:
Fixed by the PR #102!

PR #98, opened 2024-06-25
:fire: Remove unnecessary text slice

evaline-ju:
Unnecessary since in #76, text is already returned through the detectors API. We do not have to re-index to re-calculate the text

Related to: #96


gkumbhat (approved):
Looks good to me!

gkumbhat:
We can keep the sliced_codepoints function around for later. Might become handy in some aggregation processing

Issue #91, opened 2024-06-20
Failed to deserialize the JSON body into the target type: missing field `models`


pmcjr:
## Describe the bug

When calling `/api/v1/task/classification-with-text-generation` with a missing `models` object inside `guardrail_config.input` or `guardrail_config.output` objects in the payload, a `422 Failed to deserialize the JSON body` response occurs.

## Platform

Please provide details about the environment you are using, including the following:

```
fms-orchestrator image version: fms-guardrails-orchestr8:e5b72c1
```

## Observed behavior

Request body
```
{
	"inputs": "My email is",
	"model_id": "<model-id>",
	"guardrail_config": {
		"input": {
			"models": {}
		},
		"output": {
		}
	}
}
```

Response
```
Failed to deserialize the JSON body into the target type: guardrail_config.output: missing field `models` at line 9 column 3
```

## Additional context 

It happens for both `guardrail_config.output` and `guardrail_config.input`.

declark1:
Completed with https://github.com/foundation-model-stack/fms-guardrails-orchestrator/pull/99

PR #99, opened 2024-06-25
:bug: Fixed missing models field not skipping serialization on GuardrailsConfig

pmcjr:
This PR addresses [#91](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/issues/91), solved by (edit) adding validation to `models` field. 

Added a test case regarding the issue as well.


declark1:
It sounds like the objective is to:

1. Simplify the error message, i.e. remove the excess "Failed to deserialize the JSON body into the target type:" bloat and just tell the user what is missing, e.g. "guardrail_config.input: missing field models at line 6 column 3"
2. Return our custom `server::Error` type which should be returned as JSON in the following form:
    ```json
    {
	    "code": 422,
	    "details": "some error message"
    }
    ```

Axum features a [Json](https://docs.rs/axum/latest/axum/struct.Json.html) extractor that handles deserializing/serializing the requests/responses via serde behind the scenes. When deserialization fails, it returns a [JsonRejection](https://docs.rs/axum/latest/axum/extract/rejection/enum.JsonRejection.html) error. In the case of a missing field, the `JsonRejection::JsonDataError` variant is returned, and if you check the `Display` impl for it [here](https://docs.rs/axum/latest/src/axum/extract/rejection.rs.html#10-19), you can see this is where the "Failed to deserialize..." message is set. 

What we need to do is map this to `server::Error` and customize the message returned for `JsonRejection::JsonDataError` to only include the lower-level serde_json error message that tells us what is missing. There are a few ways to approach this (examples [here](https://github.com/tokio-rs/axum/tree/main/examples/customize-extractor-error)), the simplest being to leverage `axum_extra::extract::WithRejection` to transform `JsonRejection` to our custom error type `server::Error`, add a `JsonExtractorRejection` variant to `server::Error`, and then update `IntoResponse` impl for the variant.

I went ahead and gave this a try [here](https://github.com/declark1/fms-guardrails-orchestrator/commit/cad39fa1804e8ca35429537e86e3889679e3df58). The error message returned for a missing field is now:
  ```json
  {
	  "code": 422,
	  "details": "guardrail_config.input: missing field `models` at line 6 column 3"
  }
  ```

If this is sufficient, feel free to steal this. If the message needs to be customized further, it will be a more complicated task. I would recommend this approach versus converting existing fields to `Option` and adding in additional validation logic.

Btw, [here](https://github.com/tokio-rs/axum/issues/1116) is a thread with some useful information for reference.

declark1 (changes requested):
Just a couple of requested changes:
- Rebase latest changes from main
- You can drop `ValidationError::Missing` variant as it isn't being used

declark1 (approved):
LGTM. Since Evaline is out and these changes should also address her comments, will approve and merge.

Issue #53, opened 2024-05-29
Implement whole document "chunker"


evaline-ju:
## Description

As a user that may want to use various detectors, I may want to use detectors that do not require "chunking." Particularly, some detectors like regexes may just operate on entire documents.

## Discussion

For consistency with detectors that will invoke chunkers, we will implement a "default" whole document chunker that just returns a span with the start and end corresponding to an entire input document. Any detectors that do not specify a chunker will use this default one. 

For code organization purposes, we likely want to implement this not as branching logic in the orchestrator itself but as a separate chunker that can get called via some name (still implemented within the orchestrator repo, but from a deployment/user configuration POV, this name can be still be specified in the config).

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

declark1:
> For code organization purposes, we likely want to implement this not as branching logic in the orchestrator itself but as a separate chunker that can get called via some name (still implemented within the orchestrator repo, but from a deployment/user configuration POV, this name can be still be specified in the config).

Hmm, I don't think this approach will fit with the current design. We delegate to the `ChunkerClient`, which wraps the underlying `ChunkersServiceClient`s. As this "default" implementation would not be an actual grpc `ChunkersServiceClient` (which is a struct, not a trait that we can implement a `DefaultChunkerServiceClient` for), this wouldn't work.

I can propose some alternative ways to implement this, if interested.

gkumbhat:
@declark1 definitely let us know some alternatives. The idea here was to not have yet another branching logic, like `if chunker not configured` or `if chunker is default` and instead have a stock implementation of a chunker that instead of making an external calls, returns the response based on the input text directly but in the format that the "chunker" itself would do.

PR #103, opened 2024-06-26
:sparkles: Whole doc chunker

evaline-ju:
Adds a whole doc "chunker" to the chunker client that does not actually require communicating with a chunker server but instead provides the "span" for the entire document/stream. For unary this is simply providing the "length" of the doc. For streaming this requires streaming the entire stream to get the stream "length". "length" is in quotes since this calculated via codepoints and not with `.len()` for the text

Closes: #53 

May need updating once streaming/streaming updates for chunker client are updated

### DIFF
--- a/src/clients/chunker.rs
+++ b/src/clients/chunker.rs
@@ -105,14 +105,14 @@ fn request_with_model_id<T>(request: T, model_id: &str) -> Request<T> {
 
 /// Unary tokenization result of the entire doc
 fn tokenize_whole_doc(request: TokenizationTaskRequest) -> TokenizationResults {
-    let token_count = request.text.chars().count();
+    let codepoint_count = request.text.chars().count();
     TokenizationResults {
         results: vec![Token {
             start: 0,
-            end: token_count as i64,
+            end: codepoint_count as i64,
             text: request.text,
         }],
-        token_count: token_count as i64,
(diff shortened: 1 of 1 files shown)


src/clients/chunker.rs
```diff
@@ -60,13 +68,24 @@ impl ChunkerClient {
...
+        // Handle "default" separately first
+        if model_id == DEFAULT_MODEL_ID {
+            info!("Using default whole doc chunker");
+            let whole_response_stream = bidi_streaming_tokenize_whole_doc(request).await;
+            tokio::spawn(async move {
+                if let Ok(message) = whole_response_stream {
+                    let _ = tx.send(message).await;
+                }
+            });
+            return Ok(ReceiverStream::new(rx));
```
gkumbhat, line 81: I was thinking since this is essentially static if we can get away with spawning tokio task and passing along the response via channel here? Or is that not straightforward because of the type mismatch ?

gkumbhat: Or have we added this since its streaming input as well so we have to wait and thus is blocking?

gkumbhat (commented):
Approach LGTM, just left a small question around channel

gkumbhat (approved):
Looks good to me!

PR #107, opened 2024-06-28
Streaming Pipeline

declark1:
Changes:

- Split unary and streaming handlers from `orchestrator.rs` to `unary.rs` and `streaming.rs`
- Rename `GenerationClient` enum to `GenerationClientInner` and wrap in `GenerationClient` unit struct to make it mockable with faux
- Implement `tokenize`, `generate`, and `generation_stream` methods on `GenerationClient`
- Refactor and rename some existing unary task handlers
- Implement streaming pipeline (see diagram below)
- Add `DetectionAggregator` trait and placeholder `MaxProcessedIndexAggregator` implementation
- Update `NlpClient`, `ChunkerClient`, and `TgisClient` streaming methods to return client streams directly rather than mpsc channels

Streaming Pipeline Diagram (DRAFT):
O8 Streaming Diagram


src/orchestrator/streaming.rs
```diff
@@ -0,0 +1,322 @@
...
+    processor: impl DetectionStreamProcessor,
+    mut generation_stream: Pin<Box<dyn Stream<Item = ClassifiedGeneratedTextStreamResult> + Send>>,
+) -> Result<mpsc::Receiver<ClassifiedGeneratedTextStreamResult>, Error> {
+    // Create generation broadcast stream
+    let (generation_tx, generation_rx) = broadcast::channel(1024);
+
+    debug!("creating chunk broadcast streams");
+    let chunker_ids = get_chunker_ids(ctx, detectors)?;
+    // Maps chunker_id->chunk_stream
+    let chunk_streams = join_all(
```
gkumbhat, line 129: nit: can we rename `chunk_streams` to `chunk_broadcast_streams` for readability

src/clients/nlp.rs
```diff
@@ -90,20 +88,16 @@ impl NlpClient {
         &self,
         model_id: &str,
         request: ServerStreamingTextGenerationTaskRequest,
-    ) -> Result<ReceiverStream<GeneratedTextStreamResult>, Error> {
+    ) -> Result<Pin<Box<dyn Stream<Item = GeneratedTextStreamResult> + Send>>, Error> {
```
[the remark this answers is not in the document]
declark1: This is unrelated to that, it is just a simplification. Initially, I created a channel and wrapped it in a `ReceiverStream`. I am just returning the client response `Stream` directly now, i.e. the channel is unnecessary.

```rust
let (tx, rx) = mpsc::channel(128);
tokio::spawn(async move {
    while let Some(Ok(message)) = response_stream.next().await {
        let _ = tx.send(message).await;
    }
});
Ok(ReceiverStream::new(rx))
```

src/orchestrator/processors/max_processed_index.rs
```diff
@@ -0,0 +1,64 @@
...
+    models::{ClassifiedGeneratedTextStreamResult, TextGenTokenClassificationResults},
+    orchestrator::streaming::DetectionResult,
+};
+
+/// Processes detection streams applying a "max processed index" strategy.
+#[derive(Default)]
+pub struct MaxProcessedIndexProcessor {}
+
+#[async_trait]
+impl DetectionStreamProcessor for MaxProcessedIndexProcessor {
```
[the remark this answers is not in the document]
declark1: Renamed trait to `DetectionAggregator` (should probably at least have "detection" in the name) and `MaxProcessedIndexAggregator`. Let me know if you are good with this.

src/orchestrator/streaming.rs
```diff
@@ -0,0 +1,326 @@
...
+                .await;
+        } else {
+            // No HAP/PII detected
+            // Do text generation (streaming)
+            let mut generation_stream =
+                generate_stream(&ctx, model_id.clone(), input_text.clone(), params.clone()).await?;
+
+            // Do output detections (streaming)
+            let output_detectors = task.guardrails_config.output_detectors();
+            if let Some(detectors) = output_detectors {
```
gkumbhat, line 92: in case `output_detectors` are empty, it would still be `Some({})` object, so it will get matched here and go into this `if` condition, right? However, we would want it to go to `else` so that we can just process `generation_stream` ? 

Essentially, looks like we may not be following the generation when no detections are requested case ?

declark1: This is really an incorrect use of `Option`, i.e. it should really be `None` rather than `Some({})` when there are no detectors. I will update it to handle this properly, since it appears we want to do the following rather than `"output": null` or simply omitting the `output` field?

```json
"guardrail_config": {
  "output": {
     "models": {}
  }
}
```

It would probably be good to clean this up during deserialization rather than handling empty nested objects.

declark1: Pushed fix for this

src/models.rs
```diff
@@ -687,6 +688,124 @@ impl From<ExponentialDecayLengthPenalty>
     }
 }
 
+impl From<pb::fmaas::GenerationResponse> for ClassifiedGeneratedTextStreamResult {
+    fn from(value: pb::fmaas::GenerationResponse) -> Self {
+        Self {
+            generated_text: Some(value.text.clone()),
```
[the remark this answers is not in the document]
declark1: I don't believe this is feasible as we are returning a different response object. The TGIS API is stable at this point anyway, the only recent change was to add a new field to specify LoRA/other adapters.

gkumbhat (commented):
Thanks Dan for all the changes.  There are some conflicts remaining, probably arising from rebasing that we need to address, but other than that the overall changes LGTM.

declark1:
> Thanks Dan for all the changes. There are some conflicts remaining, probably arising from rebasing that we need to address, but other than that the overall changes LGTM.

Thanks. It should be good to merge now once you approve.

gkumbhat (approved):
Looks good to me!

Issue #57, opened 2024-05-31
Implement streaming result aggregation policy


evaline-ju:
## Description

Depends on #56 

As an orchestrator developer, I want to implement the aggregation policy for streamed detector results, so that I can return results to an end-user of orchestrator streaming endpoints.

## Discussion

Provide detailed discussion here

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

PR #118, opened 2024-07-10
Add max process aggregator

gkumbhat:
Supports: https://github.com/foundation-model-stack/fms-guardrails-orchestrator/issues/57

### Changes
- Add `DetectionTracker` namedtype to help with max aggregator processing
- Add logic for this processor
- Add unit test

### DIFF
--- a/src/orchestrator/aggregators/max_processed_index.rs
+++ b/src/orchestrator/aggregators/max_processed_index.rs
@@ -1,4 +1,4 @@
-use std::sync::{Arc, RwLock};
+use std::{borrow::{Borrow, BorrowMut}, collections::{BTreeMap, HashMap}, sync::{Arc, RwLock}};
 
 use async_trait::async_trait;
 use tokio::sync::mpsc;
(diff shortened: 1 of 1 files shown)


declark1:
A few initial comments:

1. Instead of creating a `AddDetectionResult` trait and implementing it for `BTreeMap<(u32, u32), (ClassifiedGeneratedTextStreamResult, usize)>`, it would be more idiomatic to create a newtype named `Tracker` (or similar) that wraps an inner container type. See below for example and I left some comments inline.

    ```rust
    // I recommend creating a type alias for span to simplify the signature. 
    // This is common in Rust code to simplify complex type definitions (clippy will often recommend this).
    type Span = (u32, u32);

    struct Tracker(BTreeMap<Span, (ClassifiedGeneratedTextStreamResult, usize)>);
    
    impl Tracker {
        pub fn new() -> Self {
            Self(BTreeMap::new())
        }

        // Note I renamed `add_detection_result` to `insert`, aligning with the container method
        pub fn insert(
            &mut self,
            span: Span,
            detection_result: Vec<TokenClassificationResult>,
            generation_result: ClassifiedGeneratedTextStreamResult,
        ) {
            todo!()
        }
    
        // Note I renamed `find_first` to `find_by_...`, more common Rust naming style for this type of thing. I named it `find_by_span_start` but something else would work too.
        // Also leveraging `iter().find()` and returning a reference to the item rather than just returning the key and having to do a subsequent `get()`
        pub fn find_by_span_start(
            &self,
            start: u32,
        ) -> Option<&(ClassifiedGeneratedTextStreamResult, usize)> {
            self.0
                .iter()
                .find(|(span, _)| span.0 == start)
                .map(|(_, value)| value)
        }
    }
    ```
2. With the above, instead of:
    ```rust
    let span = detection_tracker.find_first(processed_index);
    if span.is_some() {
        let span = span.unwrap();
        let (classified_result, num_detectors) = detection_tracker.get(&span).unwrap();
   }
    ```
    You can simplify to:
     ```rust
      if let Some((classified_result, num_detectors)) = tracker.find_by_span_start(processed_index) { 
          // ... 
      }
      ```

3. In the following code, I think you want to `extend` the `output` vec rather than replace it:
    ```rust
    self.entry((start, end))
        .and_modify(|(old_classified_stream_result, num)| {
            old_classified_stream_result
                .token_classification_results
                .output = Some(new_detection_results); // this replaces output instead of extending it
            *num += 1;
        })
        .or_insert_with(|| (classified_stream_result.clone(), 1));
    ```
4. Drop unnecessary `.clone()` on `.or_insert_with(|| (classified_stream_result.clone(), 1))`

5. For primitive types that implement `Copy`, you should use the dereference operator (`*`) for comparing a reference rather than calling `.to_owned()` and creating an owned instance:
    ```rust
    // do this:
    if *num_detectors == total_detectors {}
    
    // instead of:
    if num_detectors.to_owned() == total_detectors {} 
    ```

declark1 (changes requested):
See inline comments and suggestions

declark1 (approved):

declark1:
LGTM. 

One potential efficiency improvement -- If you don't need to keep the returned items in the BTreeMap, instead of `first_key_value`, you may want to consider [pop_first](https://doc.rust-lang.org/nightly/std/collections/struct.BTreeMap.html#method.pop_first) instead to avoid cloning the result.

Issue #119, opened 2024-07-11
Pass along include_stop_sequence to all client for generation request


gkumbhat:
## Description

Generation client, both TGIS and NLP accepts a new parameter `include_stop_sequence` .. This parameter is included in `model.rs` but it is not getting accepted by orchestrator and not getting passed to the clients. This will need to get added to allow parity with text generation.

## Acceptance Criteria

- [ ] `include_stop_sequence` added to text generation parameter in orchestrator API 
- [ ] This parameter gets passed to both TGIS and NLP clients

PR #126, opened 2024-07-24
:sparkles: Add include_stop_sequence

evaline-ju:
Surface the `include_stop_sequence` parameter as part of `text_gen_parameters` and pass along to generation clients

Closes: #119


gkumbhat (approved):
Looks good to me! Thanks for adding

Issue #136, opened 2024-07-26
Pass on server overloaded / service unavailable (503) from downstream servers to users


gkumbhat:
## Description

There is a case where detectors, text generation server or chunker may get overloaded. In those scenarios, they _may_ return proper error code and message mentioning that problem. Orchestrator would need to surface those back to the user to allow user to make decision on scaling those component.

mdevino:
Let's break this into three different issues - one for detectors, one for generation clients and one for chunkers.

declark1:
> Let's break this into three different issues - one for detectors, one for generation clients and one for chunkers.

We have now consolidated this into the same PR

PR #139, opened 2024-07-29
Detector 503 propagation

mdevino:
This addreses #136 (detectors only).

When detectors returned an error message in accordance with the API, reqwest was generating a deserialization error. This PR fixes this by mapping error responses to `DetectorError`, which can then be further propagated up the chain.


declark1 (approved):
LGTM

Issue #17, opened 2024-05-03
Add configurable request timeouts for all clients


gkumbhat:
### Description

Orchestrator is responsible to talk to several different clients. Each client can have potentially different req / response timeout requirement. This story is to add a `timeout` configuration, either common for all clients (with default) or / and allow model / detector / chunker specific timeout configuration.

### Acceptance Criteria
- [ ] Default timeout documented and added in `config/config.yaml`
- [ ] Decision for per client timeout configuration mentioned in this issue
- [ ] PR for adding this timeout support in the code merged to `main`

PR #140, opened 2024-07-29
:wrench: Allow for configuration of request timeouts per service

evaline-ju:
`timeout` in each client's service config maximizes flexibility, where each client can provide their own timeout and not be restricted to say, all detectors or all chunkers. Users can then configure each client depending potentially on resources and/or knowledge about the particular client e.g. known slow detector model

Closes: #17

### DIFF
--- a/src/clients.rs
+++ b/src/clients.rs
@@ -151,7 +151,7 @@ pub async fn create_http_clients(
             base_url.set_port(Some(port)).unwrap();
             let request_timeout = Duration::from_secs(
                 service_config
-                    .timeout
+                    .request_timeout
                     .unwrap_or(DEFAULT_REQUEST_TIMEOUT_SEC),
             );
             let mut builder = reqwest::ClientBuilder::new()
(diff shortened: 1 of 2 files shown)


src/clients.rs
```diff
@@ -149,9 +149,13 @@ pub async fn create_http_clients(
             let port = service_config.port.unwrap_or(default_port);
             let mut base_url = Url::parse(&service_config.hostname).unwrap();
             base_url.set_port(Some(port)).unwrap();
+            let request_timeout = match service_config.timeout {
```
gkumbhat, line 152: Instead of this, should we set the default via default trait ? that way we can just do `service_config.timeout` without match everywhere ?

src/config.rs
```diff
@@ -27,8 +27,13 @@ use tracing::debug;
 /// orchestrator to communicate with it
 #[derive(Clone, Debug, Default, Deserialize)]
 pub struct ServiceConfig {
+    /// Hostname for service
     pub hostname: String,
+    /// Port for service
     pub port: Option<u16>,
+    /// Timeout in seconds for request to be handled
+    pub timeout: Option<u64>,
```
gkumbhat, line 35: should we make this explicit that its "request_timeout" ?

PR #141, opened 2024-07-30
:truck: Rename orchestrator API to v2

gkumbhat:
### Changes

As per team discussion, we want to rename new APIs and put them under `v2` to better align

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -52,7 +52,8 @@ use crate::{
 };
 
 const API_PREFIX: &str = r#"/api/v1/task"#;
-const TEXT_API_PREFIX: &str = r#"/api/v1/text/task"#;
+// New orchestrator API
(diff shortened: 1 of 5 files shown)


src/server.rs
```diff
@@ -52,7 +52,8 @@ use crate::{
 };
 
 const API_PREFIX: &str = r#"/api/v1/task"#;
-const TEXT_API_PREFIX: &str = r#"/api/v1/text/task"#;
+// New orchestrator API
+const TEXT_API_PREFIX: &str = r#"/api/v2/task"#;
```
mdevino, line 56: @gkumbhat Are we dropping modality from endpoints?

gkumbhat: woops. wrong thing got removed here. We are dropping `task` not `text`.. I'll fix it. Thanks for catching!

gkumbhat: fixed!

PR #144, opened 2024-07-31
Aggregator tweaks for efficiency

declark1:
A few changes for efficiency / other cleanups:

- Read items from first & second generation message only once outside of the loop, rather than for each iteration
   - Also avoid extra clones of `ClassifiedGeneratedTextStreamResult` as extracted fields are `Copy`
- Use `.to_vec()` rather than `.iter().map(|x| x.clone()).collect()` for building `generation_responses` (more efficient)
- Change `.unwrap_or([].to_vec())` to `.unwrap_or_default()` and `.unwrap_or(&T::default()).to_owned()` to `.cloned().unwrap_or_default()`
- Drop explicit types that are inferred
- Simplify `if (input_start_index..input_end_index).contains(&0) { ... }` to `if input_start_index == 0 { ... }` (the former seems unnecessary unless I'm missing something?)

### DIFF
--- a/src/orchestrator/aggregators/max_processed_index.rs
+++ b/src/orchestrator/aggregators/max_processed_index.rs
@@ -95,7 +95,6 @@ impl DetectionAggregator for MaxProcessedIndexAggregator {
         let (result_tx, result_rx) = mpsc::channel(1024);
         tokio::spawn(async move {
             // TODO: Add chunker type
-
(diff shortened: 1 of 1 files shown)


declark1:
Closing until I add logic to await first generation message before aggregator processing (or second if `input_tokens` option is set). Will re-open once that is ready for review.

Issue #123, opened 2024-07-17
Implementation of `/api/v1/text/task/detection/context-docs`


evaline-ju:
## Description

As an orchestrator user, I want to be able to use orchestrator to call detectors that require context documents, so that I can use these detectors in my LLM pipeline

This `/api/v1/text/task/detection/context-docs` endpoint is expected to work with the [Detector API](https://foundation-model-stack.github.io/fms-guardrails-orchestrator/?urls.primaryName=Detector+API) `/text/context/doc` endpoint. 

## Discussion

The inputs to the detector are primarily the inputs to the orchestration endpoint. As this is a `detection` task (no generation), no calls to generation providers will be needed.

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

PR #147, opened 2024-07-31
Implementation of /context endpoint

mdevino:
This addresses #123. One important point to add is that `context_type` only accepts values "docs" or "url".

Scenarios tested:

* Sample request -> works just fine (base success scenario).
* Missing `context` -> returns 422
* Missing `context_type` -> returns 422
* Missing `detectors` -> returns 422
* Two detectors -> returns detections from both detectors (though the order of detections may vary)
* Threshold filtering -> only detections above the threshold are returned
* Detector returning 404 -> returns 404
* Detector returning 503 -> returns 503
* Detector returning 500 (further details are shown on logs) -> returns 500
* Detector returning 422 -> returns 422

### DIFF
--- a/src/clients/detector.rs
+++ b/src/clients/detector.rs
@@ -108,6 +108,28 @@ impl DetectorClient {
             Err(error.into())
         }
     }
+
+    /// Invokes detectors implemented with the `/api/v1/text/context/doc` endpoint
+    pub async fn detect_for_context_documents(
+        &self,
+        model_id: &str,
+        request: ContextDocsDetectionRequest,
+    ) -> Result<Vec<DetectionResult>, Error> {
+        let client = self.client(model_id)?;
+        let url = client.base_url().as_str();
+        let response = client
+            .post(url)
+            .header(DETECTOR_ID_HEADER_NAME, model_id)
+            .json(&request)
+            .send()
+            .await?;
+        if response.status() == StatusCode::OK {
+            Ok(response.json().await?)
+        } else {
+            let error = response.json::<DetectorError>().await.unwrap();
+            Err(error.into())
+        }
+    }
 }
 
 /// Request for text content analysis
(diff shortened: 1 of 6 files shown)


declark1 (commented):
Client method names should follow the names of the calling endpoints.

Currently:
`text_contents` -> `/../text/contents` (good)
`generation_detection` -> `/../text/generation`
`detect_for_context_documents` -> `/../text/context/doc`

Also, aren't some of these endpoint names being changed?

src/clients/detector.rs
```diff
@@ -108,6 +108,35 @@ impl DetectorClient {
...
+        } else {
+            let code = response.status().as_u16();
+            let error = response
+                .json::<DetectorError>()
+                .await
+                .unwrap_or(DetectorError {
+                    code,
+                    message: "".into(),
+                });
+            Err(error.into())
```
declark1, line 137: We should probably create a helper function to do this, to avoid the duplicated code in each method.

mdevino:
> Client method names should follow the names of the calling endpoints.
> 
> Currently:
> text_contents -> /../text/contents (good)
> generation_detection -> /../text/generation
> detect_for_context_documents -> /../text/context/doc
> 
> Also, aren't some of these endpoint names being changed?

I was trying to name functions using verbs as functions "do things", but I see your point.
Done!

src/clients/detector.rs
declark1: Not essential for now as we should probably refactor the Detector impl anyway as discussed in chat. We can do this then.

declark1 (approved):
LGTM

src/clients/detector.rs
mdevino: The only differences between `text_contents()`, `text_generation()` and `text_context_doc()` are:
1. The type of `request`
2. The "OK" variant type in return.
3. The detector endpoint

Let's open a separate issue and discuss some of these changes.

src/models.rs
```diff
@@ -936,6 +936,52 @@ pub struct DetectionResult {
     pub score: f64,
 }
 
+/// The request format expected in the /api/v1/text/task/generation-detection endpoint.
```
gkumbhat, line 939: This comment seems incorrect.

src/server.rs
```diff
@@ -159,6 +159,10 @@ pub async fn run(
             &format!("{}/detection/content", TEXT_API_PREFIX),
             post(detection_content),
         )
+        .route(
+            &format!("{}/detection/context-docs", TEXT_API_PREFIX),
```
gkumbhat, line 163: we are renaming this?

src/models.rs
mdevino: True, thanks for the heands up.

src/server.rs
mdevino: Yes, we agreed on renaming to `/detection/context`. I've also updated any mentions of it in the documentation.

gkumbhat (approved):
Looks good to me!

PR #148, opened 2024-07-31
:bug: Fix edge case validation for masks that reach end of inputs

evaline-ju:
Make the inputs range inclusive so that a range that includes the last character of `inputs` is legitimate

Sneaking some small indenting in

Closes: #149

### DIFF
--- a/src/models.rs
+++ b/src/models.rs
@@ -70,6 +70,9 @@ impl GuardrailsHttpRequest {
             return Err(ValidationError::Required("inputs".into()));
         }
         // Validate masks
+        // Because the masks ranges are [start, end), while applying masks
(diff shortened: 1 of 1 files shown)


src/models.rs
```diff
@@ -70,7 +70,7 @@ impl GuardrailsHttpRequest {
             return Err(ValidationError::Required("inputs".into()));
         }
         // Validate masks
-        let input_range = 0..self.inputs.len();
+        let input_range = 0..=self.inputs.len();
```
gkumbhat, line 73: Can we add a comment on top of this line explaining why `=` is required for future reader

gkumbhat (approved):
Looks good to me! Thanks for quick fix

Issue #151, opened 2024-08-05
Implementation of /api/v2/text/detection/generated 


mdevino:
## Description
As an orchestrator user, I want to be able to use orchestrator to call detectors that require generated text, so that I can use these detectors in my LLM pipeline.

This `/api/v2/text/detection/context-docs` endpoint is expected to work with the [Detector API](https://foundation-model-stack.github.io/fms-guardrails-orchestrator/?urls.primaryName=Detector+API#/default/generation_analysis_unary_handler) `/text/context/generation` endpoint.

## Discussion
The inputs to the detector are primarily the inputs to the orchestration endpoint. As this is a `detection` task (no generation), no calls to generation providers will be needed.

## Acceptance Criteria
* [ ]  Examples build against new/changed code

PR #152, opened 2024-08-05
Implement /detect/generated

mdevino:
This addresses #151.

Scenarios tested:

* Sample request -> works just fine (base success scenario).
* Missing `prompt` -> returns 422
* Missing `generated_text` -> returns 422
* Missing `detectors` -> returns 422
* Two detectors -> returns detections from both detectors (though the order of detections may vary)
* Threshold filtering -> only detections above the threshold are returned
* Detector returning 404 -> returns 404
* Detector returning 503 -> returns 503
* Detector returning 500 (further details are shown on logs) -> returns 500
* Detector returning 422 -> returns 422


src/models.rs
```diff
@@ -985,6 +985,47 @@ pub struct ContextDocsResult {
...
+}
+
+/// The response format of the /api/v2/text/detection/generated endpoint
+#[derive(Default, Debug, Clone, PartialEq, serde::Serialize, serde::Deserialize)]
+pub struct DetectionOnGenerationResult {
+    /// Detection results
+    #[serde(rename = "detections")]
+    pub detections: Vec<DetectionResult>,
+}
+
```
mdevino, line 1028: This structure has the same field as `ContextDocsResult`, but I coudn't think of a good name to reuse the already existing one. Any suggestions?

src/orchestrator/unary.rs
```diff
@@ -313,6 +313,65 @@ impl Orchestrator {
             }
         }
     }
+
+    /// Handles detections on generated textual context
```
gkumbhat, line 317:
```suggestion
    /// Handles detections on generated text (without performing generation)
```

gkumbhat (commented):
Small comment. Otherwise looks good. Thanks Mateus

src/models.rs
mdevino: Sure.

docs/api/orchestrator_openapi_0_1_0.yaml
```diff
@@ -400,6 +434,44 @@ components:
...
+        detectors:
+          type: object
+          title: Detectors
+          default: {}
+          example: 
+            generated-detection-v1-model-en: {}
+      type: object
+      required: ["generated_text", "prompt", "detectors"]
+      title: Generated-Text Detection Request
+    GeneratedTextDetectionResponse:
```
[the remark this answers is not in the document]
gkumbhat: Yeah, the reason for duplicating it here is with speculation that these differnt endpoint would end up having different parameter and response elements in future. We started with similar and simple ones, but things may expand in this direction..

src/server.rs
```diff
@@ -163,6 +164,10 @@ pub async fn run(
             &format!("{}/detection/context", TEXT_API_PREFIX),
             post(detect_context_documents),
         )
+        .route(
+            &format!("{}/detection/generated", TEXT_API_PREFIX),
```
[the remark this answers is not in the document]
mdevino: Sorry, the notification for this request didn't ring. I see that Gaurav already fixed it, thought. Thanks!

PR #156, opened 2024-08-12
Update DetectorParams to newtype wrapper

declark1:
Updates `DetectorParams` to be a newtype wrapper over `HashMap<String, serde_json::Value>`, enabling parameter flexibility across detector implementations. Includes a `threshold()` convenience method as this parameter is applied by the orchestrator.

### DIFF
--- a/src/models.rs
+++ b/src/models.rs
@@ -89,14 +89,13 @@ impl GuardrailsHttpRequest {
         if self.inputs.is_empty() {
             return Err(ValidationError::Required("inputs".into()));
         }
+        let guardrail_config = self.guardrail_config.as_ref();
         // Validate masks
         // Because the masks ranges are [start, end), while applying masks
         // will not require indexing to include the last index (i.e. len of inputs),
         // the last index is still a legitimate 'end' to provide on a mask here.
         let input_range = 0..=self.inputs.len();
-        let input_masks = self
-            .guardrail_config
-            .as_ref()
+        let input_masks = guardrail_config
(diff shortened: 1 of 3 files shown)


src/models.rs
```diff
@@ -24,10 +24,32 @@ use serde::{Deserialize, Serialize};
...
+pub struct DetectorParams(HashMap<String, serde_json::Value>);
+
+impl DetectorParams {
+    #[allow(dead_code)]
+    pub fn new() -> Self {
+        Self(HashMap::new())
+    }
+
+    pub fn threshold(&self) -> Option<f64> {
+        self.0.get("threshold").and_then(|v| v.as_f64())
```
gkumbhat, line 37: what sort of error we would get if `threshold` is provided in the request but a different data type? like if someone does `threshold: "0.0"` ?

gkumbhat (commented):
Left a small question, otherwise LG

src/models.rs
declark1: It would return `None` as it would be `Value::String` instead of `Value::Number`. `Value::as_f64()` will only return `f64` for numbers. Clients should be encouraged to use the proper data type.

declark1: Just added detector params validation, including a rule for this.

declark1:
@evaline-ju FYI I just updated this to add detector params validation to the other request objects as well:

- GenerationWithDetectionHttpRequest
- TextContentDetectionHttpRequest
- ContextDocsHttpRequest

Issue #159, opened 2024-08-13
Update evidences in detector and orchestrator APIs


evaline-ju:
## Description

As an orchestrator developer, I want to update the `evidences` in the detector API and orchestrator API, so that I can surface more evidence/explanatory info from detectors. The nesting will allow additional information like on `context` vs. `content` for the `/text/detection/context` endpoint, but at this time we should not allow infinite nesting. `name` will allow a generic description of what the evidence `value` is about

## Discussion

Each `detection` on `/text/detection/context` for the orchestrator API will look like the following:
```
[
  {
    "detection_type": "string",
    "detection": "string",
    "score": 0,
    "evidence": [
      {
        "name": "string",
        "value": "string",
        "score": 0,
        "evidence": [
          {
           "name": "string",
           "value": "string",
           "score": 0
          }
        ]
      }
    ]
  }
]
```
The `evidence` portion will be common to any result, regardless of exact `detection` (e.g. `content` results can have high level span info)

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

PR #160, opened 2024-08-14
:memo: Update evidence for detector API and context docs

evaline-ju:
This updates the detector API according to the structure as discussed in #159 . For now, `evidence` is expected for the "context docs" orchestrator endpoint i.e. `/text/detection/context`, so only the `DetectionContextDocsResponse` has been updated. As more detectors expose `evidence`, this can be added to the other orchestrator endpoints. One layer of nesting is required for now to expose the information needed by "context docs" endpoint users, but this should be the limit of nesting [more could get even more confusing].

It seems `evidence` is "uncountable" and would account for plurality, so I have updated the API references as so to not refer to "evidences"

Exposing `evidence` through the orchestrator endpoint and actual integration/testing with the "context docs" endpoint will be part of a separate issue/PR #161

### DIFF
--- a/src/clients/detector.rs
+++ b/src/clients/detector.rs
@@ -195,9 +195,9 @@ pub struct ContentAnalysisResponse {
     pub detection_type: String,
     /// Score of detection
     pub score: f64,
-    /// Optional, any applicable evidences for detection
+    /// Optional, any applicable evidence for detection
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub evidences: Option<Vec<EvidenceObj>>,
+    pub evidence: Option<Vec<EvidenceObj>>,
 }
 
 impl From<ContentAnalysisResponse> for crate::models::TokenClassificationResult {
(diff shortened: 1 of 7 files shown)


src/clients/detector.rs
```diff
@@ -154,30 +154,30 @@ impl ContentAnalysisRequest {
...
+    pub name: String,
+    /// Evidence value
+    pub value: String,
+    /// Optional, score for evidence
+    pub score: Option<f64>,
+    /// Optional, evidence on evidence value
+    // Evidence nesting should likely not go beyond this
     #[serde(skip_serializing_if = "Option::is_none")]
-    pub evidence: Option<Evidence>,
+    pub evidence: Option<Vec<Evidence>>,
```
gkumbhat, line 180: For deeper nesting, wouldn't this need to be `EvidenceObj` ?

gkumbhat (approved):
Looks good to me!

PR #174, opened 2024-08-21
:label: Make evidence value optional

evaline-ju:
Make evidence value (also in the nested evidence) optional so that detectors are not obligated to necessarily provide the evidence value, which may correspond to large texts. These large texts can either be referred to by "name", such as the context document of interest, rather than repeating the text of the context document in the "value".


gkumbhat (approved):
Looks good to me!

Issue #23, opened 2024-05-08
Add orchestrator config validation logic


declark1:
### Description

`OrchestratorConfig.validate()` is a placeholder method to implement logic that should be checked during `OrchestratorConfig::load()`. This should include anything that won't be checked during deserialization, such as ensuring that tls ref's match the name of a defined `TlsConfig`, etc.

e.g. the following config should fail validation as `caikit` isn't defined under `tls`
```yaml
# ...

chunkers:
  en_regex:
    type: sentence
    service:
      hostname: localhost
      port: 8085
      tls: caikit
tls: {}
```

**Couple of key requirements:**
1. Optional top level keys in our config: `generation`, `chunkers` , `tls` . This means that one should be able to configure orchestrator without configuring these elements.
2. We need to provide upfront error in cases when there is misconfiguration, such as: `tls` name configured in a chunker or detector or generation doesn't match with one in `tls` block of config.. Or when a `chunker` configured in `detector`, doesn't match one in `chunkers` block of config.

### Acceptance Criteria
- [ ] All needed config validation rules have been identified
- [ ] Validation rules have been implemented in `OrchestratorConfig.validate()` and return descriptive panic messages in the event of failing
- [ ] `self.validate()` is called from `OrchestratorConfig::load()`

declark1:
Yep, for the `tls_config` error, I would recommend we use `.expect("some error message")` instead of `unwrap()` to provide context.

declark1:
Closed with #176

PR #176, opened 2024-08-23
Orchestrator Service Config Validation

pscoro:
### Description

For Issue https://github.com/foundation-model-stack/fms-guardrails-orchestrator/issues/23

#### Changes
- introduced config error type variants
- made `tls`, `chunkers` and `generation` config sections optional
    - `tls` may also be provided inlined to a service config
    - `chunkers` can be omitted, in which case the default `"whole_doc_chunker"` chunker will be subsidized
    - `generation` as a whole may be omitted from configuration
- support for and error handling of optional generation client configuration
- implemented validation logic and error handling for tls config mapping
- added more config validation tests for error handling

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -180,7 +180,8 @@ impl OrchestratorConfig {
         debug!(?config, "loaded orchestrator config");
 
         Self::map_tls_configs(&mut config)?;
-        config.validate()
+        config.validate()?;
+        Ok(config)
     }
 
     async fn from_file(path: impl AsRef<Path>) -> Result<Self, Error> {
@@ -216,25 +217,25 @@ impl OrchestratorConfig {
         Ok(())
     }
 
-    fn validate(&self) -> Result<Self, Error> {
+    fn validate(&self) -> Result<(), Error> {
         if self.detectors.is_empty() {
             Err(Error::NoDetectorsConfigured)
         } else {
-            for (_, detector) in self.detectors.iter() {
-                if !self
-                    .clone()
+            for detector in self.detectors.values() {
+                // Chunker is valid
+                let valid_chunker = detector.chunker_id == DEFAULT_MODEL_ID
+                    || self
                     .chunkers
-                    .unwrap()
-                    .contains_key(&detector.chunker_id)
-                    && detector.chunker_id != DEFAULT_MODEL_ID
-                {
+                    .as_ref()
+                    .is_some_and(|chunkers| chunkers.contains_key(&detector.chunker_id));
+                if !valid_chunker {
                     return Err(Error::DetectorChunkerNotFound {
-                        detector: detector.clone().service.hostname,
+                        detector: detector.service.hostname.clone(),
                         chunker: detector.chunker_id.clone(),
                     });
                 }
             }
-            Ok(self.clone())
+            Ok(())
         }
     }
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -21,7 +21,7 @@ use std::{
 };
 
 use serde::Deserialize;
-use tracing::{debug, error};
+use tracing::{debug, error, warn};
 
 use crate::{clients::chunker::DEFAULT_MODEL_ID, server};
 
@@ -180,6 +180,13 @@ impl OrchestratorConfig {
         let config_yaml = tokio::fs::read_to_string(path).await?;
         let mut config: OrchestratorConfig = serde_yml::from_str(&config_yaml)?;
         debug!(?config, "loaded orchestrator config");
+
+        if config.generation.is_none() {
+            warn!("no generation config provided");
+        }
+        if config.chunkers.is_none() {
+            warn!("no chunker configs provided");
+        }
 
         config.apply_named_tls_configs()?;
         config.validate()?;
--- a/src/config.rs
+++ b/src/config.rs
@@ -175,20 +175,16 @@ pub struct OrchestratorConfig {
 }
 
 impl OrchestratorConfig {
-    /// Load overall orchestrator server configuration
+    /// Loads config
     pub async fn load(path: impl AsRef<Path>) -> Result<Self, Error> {
-        let mut config = Self::from_file(path).await?;
+        let config_yaml = tokio::fs::read_to_string(path).await?;
+        let mut config: OrchestratorConfig = serde_yml::from_str(&config_yaml)?;
         debug!(?config, "loaded orchestrator config");
 
         config.apply_named_tls_configs()?;
         config.validate()?;
-        Ok(config)
-    }
 
-    async fn from_file(path: impl AsRef<Path>) -> Result<Self, Error> {
-        let path = path.as_ref();
-        let s = tokio::fs::read_to_string(path).await.map_err(Error::from)?;
-        serde_yml::from_str(&s).map_err(Error::from)
+        Ok(config)
     }
 
     /// Applies named TLS configs to services.
--- a/src/config.rs
+++ b/src/config.rs
@@ -196,20 +196,21 @@ impl OrchestratorConfig {
 
     /// Applies named TLS configs to services.
     fn apply_named_tls_configs(&mut self) -> Result<(), Error> {
-        let tls_configs = &self.tls;
-        // Generation
-        if let Some(generation) = &mut self.generation {
-            apply_named_tls_config(&mut generation.service, tls_configs)?;
-        }
-        // Chunkers
-        if let Some(chunkers) = &mut self.chunkers {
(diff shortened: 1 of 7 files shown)


src/config.rs
```diff
@@ -111,55 +162,80 @@ pub struct DetectorConfig {
...
+                {
+                    return Err(Error::DetectorChunkerNotFound {
+                        detector: detector.clone().service.hostname,
+                        chunker: detector.chunker_id.clone(),
+                    });
+                }
+            }
+            Ok(self.clone())
+        }
     }
```
declark1, line 239: Can we change this to return `Result<(), Error>`? It's probably more of a style preference, but the clone can also be avoid.

Suggested code below including a few tweaks.
1. Returns `Result<(), Error>` instead of cloning self
2. `self.clone().chunkers.unwrap()` has an unnecessary clone and would panic if no chunkers are configured
3. We should check for a default `chunker_id` _or_ a valid configured `chunker_id`
4. In the error, clone `detector.service.hostname` instead of `detector`

```suggestion
    fn validate(&self) -> Result<(), Error> {
        if self.detectors.is_empty() {
            Err(Error::NoDetectorsConfigured)
        } else {
            for detector in self.detectors.values() {
                // Chunker is valid
                let valid_chunker = detector.chunker_id == DEFAULT_MODEL_ID
                    || self
                        .chunkers
                        .as_ref()
                        .is_some_and(|chunkers| chunkers.contains_key(&detector.chunker_id));
                if !valid_chunker {
                    return Err(Error::DetectorChunkerNotFound {
                        detector: detector.service.hostname.clone(),
                        chunker: detector.chunker_id.clone(),
                    });
                }
            }
            Ok(())
        }
    }
```

Update `OrchestratorConfig::load()` accordingly:
```rust
//...
config.validate()?;
Ok(config)
```

src/server.rs
```diff
@@ -85,7 +85,15 @@ pub async fn run(
     // with rustls, the hyper and tower crates [what axum is built on] had to
     // be used directly
 
-    let config = OrchestratorConfig::load(config_path).await;
+    let config = OrchestratorConfig::load(config_path)
```
declark1, line 88: Can we move the `Orchestrator::load()` call to `main.rs` and provide it to `serve()`, in lieu of `config_path`? As `main()` returns `Result<(), Box<dyn std::error::Error>`, you can then drop the conversion `From<config::Error> for server::Error`.

Suggested change for `main.rs`:
```rust
// ...
 // Launch Tokio runtime
    tokio::runtime::Builder::new_multi_thread()
        .enable_all()
        .build()
        .unwrap()
        .block_on(async {
            let config = OrchestratorConfig::load(args.config_path).await?;
            server::run(
                http_addr,
                health_http_addr,
                args.tls_cert_path,
                args.tls_key_path,
                args.tls_client_ca_cert_path,
                config,
            )
            .await?;
```

declark1 (changes requested):
Thanks @pscoro. I added a couple of comments inline. I have a couple of changes that I want to ideally add to this PR, if you don't mind me pushing a commit (if so will need permissions to write to your fork)? I didn't add them as comments as they are partially cleanups of existing code I wrote.

pscoro:
> Thanks @pscoro. I added a couple of comments inline. I have a couple of changes that I want to ideally add to this PR, if you don't mind me pushing a commit? I didn't add them as comments as they are partially cleanups of existing code I wrote.

Sounds good go right ahead

src/server.rs
pscoro: Really good suggestions thanks, resolved

pscoro:
```rust
let orchestrator = Orchestrator::new(config).await?;
```

@declark1 would it make sense to also move this line out of `Server::run` and into `main` as well? Every error after that point is strictly server related so maybe this will help clear up the errors a bit more

declark1:
> ```rust
> let orchestrator = Orchestrator::new(config).await?;
> ```
> 
> @declark1 would it make sense to also move this line out of `Server::run` and into `main` as well? Every error after that point is strictly server related so maybe this will help clear up the errors a bit more

Yep, makes sense to me. Btw, you have to sign your commits in the public github for the DCO check to pass. https://github.com/foundation-model-stack/fms-guardrails-orchestrator/pull/176/checks?check_run_id=29272541061

pscoro:
> Yep, makes sense to me. Btw, you have to sign your commits in the public github for the DCO check to pass. https://github.com/foundation-model-stack/fms-guardrails-orchestrator/pull/176/checks?check_run_id=29272541061

do we have to sign with our public github email or our ibm email?

declark1:
> > Yep, makes sense to me. Btw, you have to sign your commits in the public github for the DCO check to pass. https://github.com/foundation-model-stack/fms-guardrails-orchestrator/pull/176/checks?check_run_id=29272541061
> 
> do we have to sign with our public github email or our ibm email?

I'm not sure actually. I use my IBM one, but it probably doesn't matter.

pscoro:
commit signatures added. I think because of the rebasing it now says I committed all the commits but I preserved your authorship

pscoro:
@declark1 I pushed the last discussed change about moving orchestrator initialization out of server start up. There's a lot of follow up work that I see potential for that I have looked into, but I think for this PR I think we should wrap it up soon since every additional detail seems to want to unravel into even more fixes. Should we merge this?

List of follow up suggestions I came across while working on this and the health check issue, I have experimented with most of these on personal branches at least a bit so I can include a complexity estimate on the side of each point [/5]:
- [2/5] there is similar "try to open/resolve cert files from config" logic in `clients::create_http_clients()` and `clients::create_grpc_clients()` that can not only be factored out but can probably also be done during config validation step (if tls configs exist we may as well read the cert files and resolve the config enitrely?)
    - this will allow us to get rid of a lot of the panics in these two functions, that are otherwise a bit challenging to propagate, by handling the behaviour in an earlier more appropriate step
    - [1/5] we could then also split up `orchestrator::create_clients()` into `orchestrator::create_detector_clients()` etc. to decouple a massive function, and add error propagation. From what I can tell with this design, the main function can basically say: "we set up config, orchestrator and then server. If config fails in a predictable way, that is a fatal error. If config succeeds, then we do not expect orchestrator start up to fail (given we checked config/certs properly before hand) so this step should panic on any error currently. And finally because we separated these concerns, if the call to run the server returns erroneously we know to consider this a fatal server crash"
- [2/5] we could transform transforming a deserializable+with-optionals "unresolved" config type into a resolved config type, significantly reducing the number of optional checks/unwraps or accidental late-validation/re-validation of things that could have been caught already and certified by having a static type for resolved config (ie a type in which anything mandatory or that has a default value is no longer `Option<...>` and all the tls/cert config items are fully validated, including checking cert paths etc)
- [1/5] pretty logging of fatal errors:
image
Working around the `tracing` crate with minimal additions, we can have a graceful shutdown for application misuse (eg misconfiguration). The code to enable the above image is pretty minimal. If we handle the point above about move cert validation into the config step, then I think basically all the misuse scenarios on start up can be caught and display elegantly to the service administrator like this.
- [4/5] ExternalError infrastructure - Supposing that `clients::Error` exists to model errors encountered during or as a response from client interaction, and `server::Error` is for errors encountered during server execution that should result in an error response to the user and potential additional internal error handling:
    - Modelling errors coming in through an external protocol (http/grpc) and errors going back out can be done through shared infrastructure, I have experimented with this quite a bit and found a couple good design ideas, in particular I think a serializable/deseriazable `ExternalError` type with a code, message, and optional details object is a common model that can fit into both http and grpc error responses (see `tonic::Status::with_details`)
    
If any of these ideas seem valuable let me know, I imagine we would probably make issues for tracking, also no worries if none of these are worth investing effort into, just contributing my two cents.

declark1:
I just pushed some changes to tweak the `config::Error` variants and messages a bit and updated main to return `anyhow::Error` instead of `Box<dyn std::error::Error>`, leveraging it to show the `config::Error` `Display` impl to the user instead of `Debug`.

A few examples from testing:
```
Error: failed to read config from `/Users/dan/does_not_exist.yaml`: No such file or directory (os error 2)
Error: invalid config file: generation: missing field `provider` at line 2 column 3
Error: chunker `sentence_syntax_izumo_en_stock_chunker` not found for detector `en_syntax_slate.38m.hap`
```

I think this is good to go now, will approve and merge.

Re: your comments above, we can discuss further, but we don't need to over-engineer the error handling as there are more important items to get to.

declark1 (approved):

pscoro:
Completely understood, still getting a feel for what the priorities are, will put a pin in all of the above. Thanks for the review

PR #182, opened 2024-08-27
Fix: whole doc chunker streaming hang

### LINKED ISSUE #169: Streaming with whole doc chunker seems to hang
## Describe the bug

In configuring an output detector with the `whole_doc_chunker` and using the streaming endpoint, the call seems to hang.

## Sample Code

In config
```
detectors:
    detector_id:
        service:
            hostname: <host>
            port: 443
        chunker_id: whole_doc_chunker
        default_threshold: 0.0
```

```
curl -k 'POST' \
  'http://localhost:8033/api/v1/task/server-streaming-classification-with-text-generation' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "model_id": "model-id",
  "inputs": "Jason is a jerk. The chicken crossed the road. The social security number is 123-45-6789. His phone number is (408) 123-4567.",
  "guardrail_config": {
    "input": {
      "masks": [],
      "models": {}
    },
    "output": {
      "models": {"detector_id": {}}
    }
  },
  "text_gen_parameters": {
    "max_new_tokens": 50,
    "min_new_tokens": 50,
    "decoding_method": "SAMPLING",
    "seed": 12,
    "preserve_input_text": true,
    "input_tokens": true,
    "generated_tokens": true
  }
}
```

## Expected behavior

No hanging, results return eventually

## Observed behavior

Hanging call

## Additional context

Unary call with `classification-with-text-generation` seems to return expected results

declark1:
Hanging was observed while creating the chunk broadcast streams using the "whole doc chunker". This is because the call to `tokenize_whole_doc_stream().await` in `ChunkerClient::bidi_streaming_tokenization_task_predict()` does not resolve until the input (generation) stream is closed.

The "whole doc chunker" is not practical for streaming as it collects the entire input stream and returns a single response, i.e. there is no benefit of streaming at all. However, this PR at least resolves the hanging issue by spawning a tokio task to await the output of `tokenize_whole_doc_stream()` and sending it to a channel wrapped in a `ReceiverStream`.

Closes #169


gkumbhat (approved):
Looks good to me!

Issue #163, opened 2024-08-16
Perform client health checks on start-up to further validate configuration


declark1:
## Description

As an extension of config validation, we should ensure that all clients (generation, detectors, chunkers) are healthy on start-up to catch any misconfigured endpoints or certs.

## Discussion

Currently, we create clients in `Orchestrator::new()` [here](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/main/src/orchestrator.rs#L60). Instead, we should create them after loading/validating the config [here](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/main/src/server.rs#L89), perform health checks, and inject them into `Orchestrator::new()` as a dependency.

We should panic in the event of a failed health check so that the configuration/other issue can be addressed.

The health checks can be triggered inside of `create_clients()` or as a separate function. To enable this functionality, `health` methods need to be implemented for each client. The gRPC clients (tgis, nlp, chunker) may not have the gRPC health service implemented (check and confirm). If not, let's discuss further offline as they may implement a separate standalone HTTP service for health checks.

## Acceptance Criteria

- [ ] Health methods are implemented for each client type
- [ ] Health checks are performed on start-up when clients are created
- [ ] Failed health checks panic with a descriptive error message indicating the specific client that is failing

pscoro:
> but I’m concerned about the strategy of bringing down orchestrator on unhealthy/unready clients

Hey Evaline, in my PR I kept "run a health probe and aggregate the result" and "what do we do because of this result" as separate items. In theory you can run the probe whenever without it causing any effect on the orchestrators liveliness. The strategy of bringing down the orchestrator only happens during the "start up check" immediately after starting the orchestrator. If the readiness probe were to be triggered by a user visiting the endpoint this will have no effect on the orchestrator's liveliness. Similarly, if we are reading the results of the readiness probe at any other point in the orchestrator, we can choose whether failing health probe can imply fatal behaviour or not. At the moment it was just assumed that there needs to be a probe, exposed through an endpoint, and on start up if it fails the orchestrator is particularly likely to be misconfigured. maybe there can be a flag to bypass start up health checks (say we already know clients will fail, but we want to boot up the orchestrator anyway). Let me know what you think/if i misunderstood anything.

pscoro:
Thanks, these are some good points.

[...]

PR #183, opened 2024-08-28
Orchestrator HTTP/gRPC health checks and health readiness probe

pscoro:
### Issue
https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1222
OSS: https://github.com/foundation-model-stack/fms-guardrails-orchestrator/issues/163

### Changes
- Health checks supported by orchestrator for both gRPC and HTTP:
    - any client's health status can be `READY`, `NOT_READY`, or `UNKNOWN`
    - gRPC: Introduce health check service proto (using the v1 health check proto officially provided by gRPC: https://github.com/grpc/grpc/blob/master/src/proto/grpc/health/v1/health.proto, introduce corresponding health client for each grpc client.
    - HTTP: uses a minimal health contract
        - health endpoint: `/health`
        - empty `200 OK` -> `READY`
        - non-empty `200 OK` -> try parse health_status (e.g. `{ "health_status": "UNKNOWN" }`), if HTTP response body is not empty this is required. Optionally a `reason` can also be provided `{ "health_status": "NOT_READY", "response_code": "HTTP 200 OK", "reason": "some reason" }`. This was done in order to allow http endpoints to potentially use `200 OK` for any successful response even if the response is to say `NOT_READY` or `UNKNOWN`, so if the client just wants to return `READY`, a body can/should be omitted. If a 200 OK response is to be considered `READY` and the response does contains a body, it must have `"health_status": "READY"`, but if `"reason": "..."` is specified it will be ignored as this is meant as an additional field for unhealthy reporting
        - `503 SERVICE UNAVAILABLE` and all other `5xx` HTTP codes are treated as `NOT_READY` because of the implication that a `5xx` error is the fault of the server.
        - `404 NOT FOUND` and all other `4xx` HTTP codes are treated as `UNNKNOWN` because of the implication that a `4xx` error is the fault of the client, and so the most we know is that we do not know the health status and that this could likely be caused by misconfiguration/misuse rather than a server issue
- Implement readiness probe for the actual client "handlers" (generation, detection, chunker) to collect the health check results of the clients
- expose readiness probe through orchestrator and introduce a health status cache in orchestrator for clients
    - introduce `/ready` server endpoint (as separate from `/health` which is a superficial orchestrator server check)
    - `/ready` takes `?probe=true/false` query param with default `false`, when `true`, a readiness probe and update of health status cache is triggered, otherwise just read cache.
    - returns a serialization of `HealthStatusCache`, showing status for each client in detectors, chunkers, generation. 
        -  `READY` clients' statuses are simplified `"some_client_id": "READY"`, while `NOT_READY` and `UNKNOWN` statuses include more information (e.g. `{ "health_status": "UNKNOWN", "response_code": "HTTP 404 Not Found", "reason": "HTTP status client error (404 Not Found) for url (http://no.url.provided.local/): { \"message\": \"service not found\" }" }`)
    - readiness probe is triggered on orchestrator start up and (only at startup) `NOT_READY` or `UNKNOWN` statuses are considered fatal for the server.
- added unit tests mocking responses from HTTP and gRPC clients, testing handling of the different successful and unsuccessful cases.
- Included a commit that updates the OpenAPI spec, if this is wanted.


declark1 (commented):
Looks pretty good to me. I left some comments inline, mostly nits.

I noticed the calls to the underlying clients in the `HealthProbe` impl aren't performed concurrently, which is probably not an issue for the generation clients as I anticipate those to actually be single "router" clients. If we have a lot of chunkers and detectors, we may want to consider doing so, depending on the probe frequency.

src/health.rs
```diff
@@ -0,0 +1,339 @@
...
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "overall health status: {}\n{}",
+            self.health_status, self.services
+        )
+    }
+}
+
+impl Serialize for HealthCheckResult {
```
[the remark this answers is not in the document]
pscoro: I didnt want to clutter a response that contains mostly `READY` responses with full objects for each client and wanted a shorthand string status:

What the response should look like:
```json
{
    "health_status": "NOT_READY",
    "services": {
        "generation": {
            "model_id_1": "READY",
            "model_id_2": "READY",
            "model_id_3": {
                "health_status": "NOT_READY",
                "response_code": "HTTP 503 Service Unavailable",
                "reason": "HTTP status server error (503 Service Unavailable) for url (http://no.url.provided.local/): { \"message\": \"some error message\" }"
            },
            ...
        },
        "detectors": { ... },
        "chunkers": { ... },
    }
}
```

pscoro:
Left an explanation about one of the comments and resolved nits from all the others

pscoro:
Update to start up behaviour:

Based on discussion regarding requirements for start up validation, I've updated the PR so that the client health statuses are not considered part of a validation process on orchestrator start-up. The orchestrator still performs the client health probe on start-up, but only so that it can have the results ready in cache (TODO: do we want a flag for this?). The client health probe does not provide an overall status, as it now only relays the health/readiness of clients and makes no inference about the orchestrator itself. The client services are not dependencies of the orchestrator. Failing client health checks will not impede the orchestrator's liveliness during start-up or after any other re-probing. The client health probe is now exposed by `/info` instead of `/ready`, a `?probe=true` can still be provided. Below is an example of the logging output on orchestrator startup for a config that contains clients that fail to pass health checks:

image

pscoro:
Issue arising from a conflict with the recently merged introduction of integration tests by @mdevino: https://github.com/foundation-model-stack/fms-guardrails-orchestrator/pull/188

Summary of problem and discussion so far: 

- Background info - a merge conflict for health probe PR caused by recent introductory integration test PR causes the questions about mocking w.r.t. integration tests to need addressing

- Problematic detail - we already use some mocking for unit tests with e.g. #[cfg_attr(test, faux::create, derive(Default))] , however the built-in build configuration condition test is purposely not enabled by rust for integration tests (tests in tests/)
Why not? - Taking a step back, the point of integration tests is to test the integration of all modules of the code together, as they would actually be used. Mocking is generally a discouraged practice for integration tests as you can not test that all the parts of a system are working together when some of the parts are mocked. Rust adheres to this premise and thats why conditional build configurations for integration tests are not enabled by the test feature.

- Short term fix - when I found this issue on Friday I made a temporary solution where a custom cargo feature "mock" acting in place of test is made required by the integration test instead, and the build configuration attributes in the code were modified to #[cfg_attr(any(test, feature = "mock"), faux::create, derive(Default))]. Building off the previous point though, perhaps we want to re-evaluate our approach to integration testing to avoid mocking parts of the system

- Speculation about a real fix - IMO I think a more correct approach to integration testing should not mock any orchestrator code, instead it can mock the client services. Maybe something like an orchestrator config file for very simple localhost instances of client services (just simple scripts mocking the client service behaviour) that return dummy responses without any additional logic

- What about functional tests? - The main difference in purpose I see between our functional tests and the proposal for integration tests is that functional tests are for validating the business logic while integration tests are testing that the components of the system interact as expected. As such, functional tests are black box tests that validate the response contains correct logic given a request, while an integration test is a (white box?) test that ensure a request is handled and responded to as expected without caring about the contents of the dummy request/responses used

As it stands, neither @mdevino or I have a clear direction about what the correct approach is and believe this should be discussed further with the team.

pscoro:
cc @declark1 I resolved the failing chunker health check issue (caikit expects an empty string to be passed for the `service_id` of a `HealthCheckRequest` and specifying the chunker service id was causing the failure) and addressed the other discovered integration test issue. This PR is now ready for final review

src/health.rs
```diff
@@ -0,0 +1,307 @@
...
+        matches!(self.health_status, HealthStatus::Ready)
+    }
+
+    pub fn is_not_ready(&self) -> bool {
+        matches!(self.health_status, HealthStatus::NotReady)
+    }
+
+    pub fn is_unknown(&self) -> bool {
+        matches!(self.health_status, HealthStatus::Unknown)
+    }
```
declark1, line 112: None of these methods are actually used, so they should be dropped

src/health.rs
```diff
@@ -0,0 +1,307 @@
...
+                    response.status
+                );
+                Some(format!(
+                    "Unexpected gRPC health check response status: {}",
+                    response.status
+                ))
+            }
+        }
+    }
+}
```
declark1, line 140: These descriptions feel redundant, aren't the status strings such as UNKNOWN, NOT_SERVING, etc.  self-explanatory?

src/clients.rs
```diff
@@ -125,20 +128,123 @@ impl From<tonic::Status> for Error {
...
     }
 
     pub fn base_url(&self) -> &Url {
         &self.base_url
     }
+
+    /// This is sectioned off to allow for testing.
+    pub(super) async fn http_response_to_health_check_result(
+        res: Result<Response, reqwest::Error>,
+    ) -> HealthCheckResult {
```
declark1, line 170: All of the logic in this method feels unnecessary IMO. I think we can simply consider a non-200 status code on a health endpoint to be unhealthy? 

Also, I think `HealthStatus::NotReady` should be renamed to `HealthStatus::Unhealthy`, as we are checking health, not readiness, and `HealthStatus::Unknown` should be dropped as a service is either healthy or it isn't, why introduce the concept of "unknown" here? Unknown essentially means unhealthy in this context.

declark1 (changes requested):
I left a couple of comments inline as I think some of this can be simplified and unused methods should be dropped. Also, I suggest that we rename `HealthStatus::NotReady` to `HealthStatus::Unhealthy` (we are checking health, not readiness), and I think `HealthStatus::Unknown` should be dropped as a service is either healthy or it isn't; unknown essentially means unhealthy in this context.

src/health.rs
pscoro: Regardless of whether we use `HEALTHY | UNHEALTHY | UNKNOWN` or just `HEALTHY | UNHEALTHY`, we can preserve the original gRPC status in non-healthy cases as part of the reason message for debugging, where gRPC health status is of `READY | NOT_READY | SERVICE_UNKNOWN | UNKNOWN`. I think this is something nice to preserve, although if these are too wordy maybe simpler messages like just `from gRPC serving_status: SERVICE_UNKNOWN`, omitting the redundant `Service's health is currently unknown`?

declark1 (approved):
Per discussion, we will merge this as-is and later open a separate PR to make health checks configurable/optional.

PR #185, opened 2024-09-05
:label: Update based on open-sourced types

evaline-ju:
Chunkers data models were open-sourced in caikit in https://github.com/caikit/caikit/releases/tag/v0.27.2 - the main "breaking" change is the field number update

Should be considered a breaking change for orchestrator (at least a minor semver update prior to 1.0.0) and should be coordinated with updates of any deployed chunker servers


protos/caikit_data_model_nlp.proto
```diff
@@ -27,6 +27,16 @@
 
  /*-- MESSAGES ----------------------------------------------------------------*/
 
+  message ChunkerTokenizationStreamResult {
```
gkumbhat, line 30: nit: Can we reference caikit proto / DM here, so that this doesn't look arbitrary and also explain why `input_start_index` etc are starting at `20`

gkumbhat (approved):
Looks good to me!

PR #186, opened 2024-09-09
:bulb: Add config description

evaline-ju:
Provide more inline description for orchestrator config sections for users


gkumbhat (approved):
This is very helpful. Thank you

Issue #112, opened 2024-07-08
Response's classification results ordering by span behaving inconsistently


pmcjr:
## Describe the bug

When calling `/api/v1/task/classification-with-text-generation` with multiple models present in `guardrail_config.input` objects in the payload, response's  `token_classification_results.input` ordering by span are not consistent.

## Platform

Please provide details about the environment you are using, including the following:

```
fms-orchestrator image version: fms-guardrails-orchestr8:ee4dd81
```

### Request body for both calls
```
{
  "inputs": "I hate cats they are stupid. I love dogs love@dogs.com. Rabbits are pretty but",
  "model_id": "bloom-560m",
  "guardrail_config": {
    "input": {
      "models": {
        "en_syntax_slate.38m.hap": {
          "threshold": 0.5
        },
        "en_syntax_rbr_pii": {
          "threshold": 0.8
        }
      }
    },
    "output": {
      "models": {}
    }
  },
  "text_gen_parameters": {
    "preserve_input_text": true,
    "max_new_tokens": 99,
    "min_new_tokens": 1,
    "truncate_input_tokens": 500,
    "decoding_method": "SAMPLING",
    "top_k": 2,
    "top_p": 0.9,
    "typical_p": 0.5,
    "temperature": 0.8,
    "seed": 42,
    "repetition_penalty": 2,
    "max_time": 0,
    "stop_sequences": [
      "42"
    ]
  }
}
```

#### Response 1 
![response_1](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/assets/7291154/f7354509-4891-4e72-bb57-9fdd346e6fae)

#### Response 2 
![response_2](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/assets/7291154/8a7e1553-8ae5-4837-a23b-ab4ac104f10c)

PR #190, opened 2024-09-13
:sparkles: Detection result ordering by start span

evaline-ju:
Due to parallel detection calls, there is currently some unpredictability on detection result ordering
- For unary responses (including all input detection), order detections by start span
- Within each chunk for streaming output, order detections by start span

Note: This only applies for detection `/contents` and generation endpoints that invoke `/contents`. Other detection results currently do not have span information and are not ordered

Closes: #112


gkumbhat (approved):
Looks good to me!

Issue #189, opened 2024-09-09
Add testing framework for integration tests


mdevino:
## Description

As aorchestrator developer, I want to have a testing framework to, so that I can write integration tests.

## Discussion

We use [Axum](https://github.com/tokio-rs/axum) as the web framework for orchestrator, and it does not support a test client natively. To make tests simpler, we could add a framework, such as [axum-test](https://github.com/JosephLenton/axum-test) to assist on writing tests.

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code

mdevino:
Address by #188.

PR #200, opened 2024-09-18
Prepare integration tests for stateful request handling without mocking orchestrator

pscoro:
Part of/follow-up addition to: https://github.com/foundation-model-stack/fms-guardrails-orchestrator/issues/189

As a follow up to discussions regarding integration testing approach and the conflicts with the [health check probe PR](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/pull/183), I propose the following pre-emptive changes that better belong in a separate PR than as a part of the health check probe PR.

Changes in this PR:
- prepares the integration tests for testing flows that require a stateful server/real orchestrator instance without mocking
- introduces test config file used by test instance of orchestrator - note that we do not currently mock client services and we will need to for future additional integration tests but for the health endpoint, and the to-be-added info endpoint from the health check PR, this wont be a concern since we can just do a superficial test for whether the result is `200 OK` since we determined that the orchestrator should not treat client services as dependencies. Once client service mocking is added we can improve health endpoints testing and introduce tests for more complex endpoints/requests
- preemptively added a server state to the health router (health check PR will necessitate this) - added this in this PR so an orchestrator needs to be set up for the integration tests
- using static lazy initialization (asynchronously with `tokio::sync::OnceCell`) to only need to set up an initial orchestrator state once for all integration tests
- added `tracing-test` so that tests can make use of `tracing` for logging, useful so that we can use `debug!` instead of `println!` in tests, nicer output and also configurable, but not strictly needed. This is a small, well-trusted, apache licenced crate built around `tracing`.

Following this PR, the health check probe PR can seamlessly integrate (albeit at the moment only superficial) testing of the `/info` endpoint without needing to add any temporary hacks or revert anything regarding integration testing.


mdevino (commented):
Looks good to me. Just left a comment.

src/server.rs
```diff
@@ -84,10 +90,10 @@ pub async fn run(
     // with rustls, the hyper and tower crates [what axum is built on] had to
     // be used directly
 
-    let shared_state = Arc::new(ServerState { orchestrator });
+    let shared_state = Arc::new(ServerState::new(orchestrator));
```
mdevino, line 93: Hey @pscoro, just for my own understanding, what's the need of the `new()` method here instead of what we had?

pscoro: This was just a small but relevant improvement, since we are using server state in multiple places. Dont have much additional reasoning for this change.

src/server.rs
```diff
@@ -263,8 +269,10 @@ pub async fn run(
     Ok(())
 }
 
-pub fn get_health_app() -> Router {
-    let health_app: Router = Router::new().route("/health", get(health));
+pub fn get_health_app(state: Arc<ServerState>) -> Router {
+    let health_app: Router = Router::new()
+        .route("/health", get(health))
+        .with_state(state);
     health_app
```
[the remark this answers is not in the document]
pscoro: good catch, will commit this

mdevino (approved):
LGTM!

config/test.config.yaml
```diff
@@ -0,0 +1,18 @@
+generation:
```
gkumbhat, line 1: This can become confusing, can we move this to `tests/` folder ?

tests/canary_test.rs
```diff
@@ -1,11 +1,42 @@
...
+    let orchestrator = Orchestrator::new(config).await.unwrap();
+    Arc::new(ServerState::new(orchestrator))
+}
 
 /// Checks if the health endpoint is working
+/// NOTE: We do not currently mock client services yet, so this test is
+/// superficially testing the client health endpoints on the orchestrator is accessible
+/// and when the orchestrator is running (healthy) all the health endpoints return 200 OK.
+/// This will happen even if the client services or their health endpoints are not found.
+#[traced_test]
```
gkumbhat, line 32: what is the need for traced_test (and adding this library) ?

Cargo.toml
```diff
@@ -43,6 +43,7 @@ url = "2.5.2"
 uuid = { version = "1.10.0", features = ["v4", "fast-rng"] }
 async-trait = "0.1.81"
 async-stream = "0.3.5"
+tracing-test = "0.2.5"
```
gkumbhat, line 46: can we move this to `dev-dependencies` ?

gkumbhat (changes requested):

tests/canary_test.rs
pscoro: This was definitely not a strong need by any means so I'll let you have the final say on whether we want it or not. It just enabled testing to use tracing but in retrospect the additional value is perhaps too minimal. Let me know if you want to use this (so we can use `debug!`) or if you prefer to just `println!`. I originally added this as I was trying to decide between leaving the `println!`s in the PR or not, as I found them useful for debugging but they require some additional work (json parsing, pretty printing) that made it feel worth it to leave the `println!`/`debug!`s behind so that someone debugging in the future doesn't need to re-figure out how to pretty print the response. And then I thought if I'm leaving behind debug prints, maybe I should make them explicitly `debug!` with the potential additional benefit of having log level configurable.

If you do want to keep `tracing-test` around, I will move it to dev-dependencies, but I also don't need any convincing to just remove it.

gkumbhat: make sense. we should not use println, so lets keep `tracing-test` and move it to `dev-dependencies`. Thanks

pscoro:
All nits should be resolved, determined to keep `tracing-test`, moved to `dev-dependencies`, ready for final review

gkumbhat (approved):
Looks good to me!

PR #201, opened 2024-09-18
:recycle: Re-order endpoints and add tags

evaline-ju:
Mainly re-ordering endpoints to group the "text generation" ones together, "detection tasks" one together, and add tags

Goal here is to make endpoints easier to navigate for any devs trying to add more

### DIFF
--- a/docs/api/orchestrator_openapi_0_1_0.yaml
+++ b/docs/api/orchestrator_openapi_0_1_0.yaml
@@ -3,17 +3,17 @@ info:
   title: FMS Orchestrator API
   version: 0.1.0
 tags:
-    - name: Text Generation with detections
+    - name: Task - Text Generation, with detection
       description: Detections on text generation model input and/or output
-    - name: Detection tasks
+    - name: Task - Detection
       description: Standalone detections
-    - name: Chat Completions with detections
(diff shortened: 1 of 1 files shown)


docs/api/orchestrator_openapi_0_1_0.yaml
```diff
@@ -2,11 +2,18 @@ openapi: 3.0.1
 info:
   title: FMS Orchestrator API
   version: 0.1.0
+tags:
+    - name: Text Generation with detections
+      description: Detections on text generation model input and/or output
```
gkumbhat, line 7: nit: Can we specify "tasks" more explicitely here? 

Something like: `Task - Generation <br /> Description - Detections on text generation model input and/or output`

docs/api/openai_openapi.yaml
gkumbhat, line 1: This one got pushed accidentally ?

docs/api/orchestrator_openapi_0_1_0.yaml
gkumbhat: not quite, but this works as well. Thanks

gkumbhat (approved):
Looks good to me!

Issue #202, opened 2024-09-18
Add to README description of setup needed to run orchestrator


evaline-ju:
## Description

As a orchestrator user, I want it to be clear what setup needs to be done to potentially try out the orchestrator locally.

## Discussion

Add to README links to config, notes on setup of servers

## Acceptance Criteria

- [ ] READMEs are updated

PR #204, opened 2024-09-19
:memo: Document APIs and server config in README

evaline-ju:
Note that server connections are needed for orchestrator to run an explicit part of README

For: #202


gkumbhat (approved):
Looks good to me!

Issue #165, opened 2024-08-19
Add detector_type to configure different detector endpoints


gkumbhat:
## Description

As we are starting to add multiple detectors of different types, we need a way to easily map and differentiate from one another, instead of relying the full endpoint.  For this purpose, we want to introduce a `detector_type` ENUM in orchestrator that will map 1:1 with 1 type of detector corresponding to 1 endpoint (suffix). This will help and get used in:
1. Configuring different detector (in `config.yaml`) and replace the need to provide full url
2. Validation, if a model is of particular `detector_type` and it is being called for a different endpoint, then we can easily throw a validation error to user
3. (Future use): Allow future use-cases of complex API workflows where different detector types are supposed to automatically work at different part of the API lifecycle.

## Acceptance Criteria

- [ ] Add ADR documenting concept of detector types
- [ ] Implement detector type in orchestrator  (need not implement validation logic, point 2 and 3 above)

PR #205, opened 2024-09-20
Add detector type to config

mdevino:
This addresses #165.

I'm opening this as I'd like to get some feedback on what I've done so far.

Missing tasks:
* [x] Add ADR
* [x] Either use generics or create a helper function for `text_contents(), `text_generation()` and `text_context_doc()` - as [faux has a limitation when dealing with generics](https://github.com/nrxus/faux/issues/18), I've rolled back this change and will work on it in a different PR.

### DIFF
--- a/src/clients.rs
+++ b/src/clients.rs
@@ -27,7 +27,7 @@ use tracing::error;
 use url::Url;
 
 use crate::{
-    config::{ServiceConfig, Tls},
+    config::{DetectorConfig, DetectorType, ServiceConfig, Tls},
     health::{HealthCheck, HealthCheckResult, HealthStatus, OptionalHealthCheckResponseBody},
 };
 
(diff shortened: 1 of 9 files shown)


src/clients.rs
```diff
@@ -129,16 +129,31 @@ impl From<tonic::Status> for Error {
 pub struct HttpClient {
     base_url: Url,
     client: reqwest::Client,
+    detector_type: DetectorType,
```
mdevino, line 132: Is this okay, or should we have a "subclass" for detector clients?

src/clients.rs
```diff
@@ -129,16 +129,31 @@ impl From<tonic::Status> for Error {
...
+            client,
+            detector_type,
+        }
     }
 
     pub fn base_url(&self) -> &Url {
         &self.base_url
     }
+
+    pub fn endpoint(&self) -> String {
```
mdevino, line 148: Would it be better to return `String`, `&'static str` or `&Url` here?

src/clients.rs
```diff
@@ -129,16 +129,31 @@ impl From<tonic::Status> for Error {
...
+        let path = match self.detector_type {
+            DetectorType::Content => "/api/v1/text/contents",
+            DetectorType::Generated => "/api/v1/text/generation",
+            DetectorType::Chat => "/api/v1/text/context/chat",
+            DetectorType::Context => "/api/v1/text/context/doc",
+        };
+        self.base_url.join(path).unwrap().to_string()
+    }
```
mdevino, line 156: Would it be better to return `String`, `&'static str` or `&Url` here?

src/config.rs
```diff
@@ -119,6 +119,17 @@ pub struct ChunkerConfig {
     pub service: ServiceConfig,
 }
 
+#[derive(Clone, Copy, Debug, Deserialize, Default)]
+#[cfg_attr(test, derive(PartialEq))]
+#[serde(rename_all = "lowercase")]
+pub enum DetectorType {
+    #[default]
```
mdevino, line 126: Note: even though this is set as default, "type" is mandatory on the YAML.

declark1 (commented):
We should not add detector concepts, e.g. `detector_type` and `endpoint` with detector-specific endpoints to `HttpClient`. This is meant to be a reusable abstraction for any HTTP API, e.g. we will be using it for the upcoming OpenAI chat & chat completions implementation.

mdevino:
> We should not add detector concepts, e.g. `detector_type` and `endpoint` with detector-specific endpoints to `HttpClient`. This is meant to be a reusable abstraction for any HTTP API, e.g. we will be using it for the upcoming OpenAI chat & chat completions implementation.

@declark1 Agreed, but not quite sure on how to go about this in a _Rusty_ way. Should I move these out to a trait, or create a whole new struct?

declark1:
Closing this as we are consolidating client-related changes into #220

Issue #167, opened 2024-08-19
Feature: Support Chat API


gkumbhat:
### Use-case

End-to-end guardrails experience exposed on openAI chat API.  This means, a user can use similar API as openAI chat and provide a list of detectors they want to run. Behind the scene, orchestrator will automatically run the detector at appropriate time in the lifecycle of the request (based on type of detector).

### Config requirements
Option 1:
- Using same `generation` section and potentially add multiple providers there (one of them being chat server)

Option 2:
-  Add a new section for chat

## Acceptance Criteria
Design above interaction and document design and implementation changes we need to do to support this.

declark1:
For reference, OpenAI Chat/Completions API implementation for TGIS router (not merged): https://github.com/IBM/text-generation-router/tree/openai-api

PR #209, opened 2024-09-25
:memo::sparkles: Add API for detections on chat completions and ADR

evaline-ju:
In this PR:
- ADR for chat completions API decisions
- Chat completions with detections API with request and response objects
     - Since this is meant to build directly on the [OpenAI chat completions API](https://platform.openai.com/docs/api-reference/chat), the OpenAI API is referenced directly in the schema. The additive portions of `detectors` and `detections` and `warnings` respectively will be more obvious
     - Some refactor was done to reference array elements as part of the new response object, since this endpoint could support a variety of detectors and thus detection results
     - Classification objects were just moved as is to provide some grouping - no updates were made

For: #167

### DIFF
--- a/docs/architecture/adrs/005-chat-completion-support.md
+++ b/docs/architecture/adrs/005-chat-completion-support.md
@@ -27,7 +27,7 @@ This is particular to the [`choices` array](https://platform.openai.com/docs/api
 2. Each choice is generated with the same input.
 
 #### Response
-- When output detectors are requested, if the LLM (model used to generate chat completions) doesn't generate `content` in any `message` in `choices`, then we will ignore the `choices` and return a warning about `detections.output`.
+- When output detectors are requested, if the LLM (model used to generate chat completions) doesn't generate `content` in any `message` in `choices`, then we will ignore the `choices` and return a warning about `detections.output`. This was written based on detectors at this time designed to work on `content`, such as text detectors. This may change if new detectors work on tool calls, for example.
 - We will run all detectors included in `detectors.output` of the user request on all `choices` (that have `message.content`), unless there are certain detectors that are supposed to be run across choices.
 - We will order all detections that include span by `start` and move all the other ones towards the end (by not ordering them in any particular order), but grouping the remaining ones based on `detector_id`.
     - An alternate consideration made here was to not order the detections and just include `detector_id`. However, since detectors are called in parallel, just collecting results may lead to different ordering each call. This ordering will allow processing of detections with spans in order by `start`.
@@ -37,9 +37,10 @@ This is particular to the [`choices` array](https://platform.openai.com/docs/api
 
 The streaming response with guardrails is built off the [chat completion chunk responses](https://platform.openai.com/docs/api-reference/chat/streaming). This applies when both streaming and output detectors are requested.
 - Each choice's `delta.content`, if present, will be the `content` used for `detections`. Like the unary case, if there is no `content` overall, a warning will be returned.
-- Since detector use will cause content to be "chunked" (e.g. collected into sentences for sentence detectors), to prevent ambiguity, the `delta.role` will be provided on each stream event, and `delta.content` will include the "chunked" text. This will apply to each `choice` in `choices`.
+- Since detector use will cause content to be "chunked" (e.g. collected into sentences for sentence detectors), to prevent ambiguity, the `delta.role` will be provided on each stream event, and `delta.content` will include the "chunked" text. This will apply to each `choice` in `choices`. Like previous [text generation cases](./002-streaming-response-aggregation.md#end-user-experience), this API will also only return fully processed chunks, to ensure that users receive text back only when all detectors have been run.
 - A `detections` block will be added to each stream event. Each processed `choice` will be noted with the inclusion of the choice's index in the original `choices` array by a `choice_index` field, even if there are no particular `results` from running the output detectors.
 - For output detectors that run on the contents of an entire output stream, results will be included as `detections` on the second-last event (last one before `[DONE]`) of the stream, regardless of presence of `usage`.
+    - Note here that overall `usage` is returned as-is to the user, just adding `detections`. Likewise, any other optional response parameters will be unchanged.
     - An alternate consideration made here was the currently implemented aggregation strategy, where if the user requested any output detectors that required the entire output, the stream events with detection results would only be returned all together. The problem here was the results would then not really be "streamed" back.
 
 
@@ -53,7 +54,7 @@ This part gives a brief overview of how we would expect a user to access informa
 - We will add a `warnings` element to show cases where the response is not 4xx, but there are issues in processing.
 
 ## Consequences
-- Only additive changes to the OpenAI chat completions results will be made in the unary case. In the streaming case, streamed events may need to have updated text in `delta.content`. This can affect token counts per event in `usage`.
+- Only additive changes to the OpenAI chat completions results will be made in the unary case. In the streaming case, streamed events may need to have updated text in `delta.content`. This can affect token counts per event in `usage`, as well as what `tokens` are included for `logprobs.content` or similar fields.
 - Validation and error handling will have to be provided for the cases where there is not expected `content` to run detectors on, raising warnings to the user when appropriate.
 - Since the results of various detectors can be included as the results of each `choice`, the inclusion of `detector_id` in each result will help distinguish which detector produced which result.
 - Results from input detectors will include `message_index` to indicate which message was processed. In cases where only the last message of the chat history may be processed to limit resource usage, this will help make it obvious to the user which message was processed.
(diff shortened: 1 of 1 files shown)


mdevino:
image

I cloned the branch to my repo and the UI looks like this. Maybe we should keep a single header to be consistent with the other sections?

docs/architecture/adrs/005-chat-completion-support.md
```diff
@@ -0,0 +1,74 @@
...
+- Different chat completions server solutions like [vllm](https://docs.vllm.ai/en/stable/index.html) may provide [extra parameters](https://docs.vllm.ai/en/stable/serving/openai_compatible_server.html#extra-parameters-for-chat-api). For the orchestrator API to remain more agnostic of particular server implementation, the orchestrator can pass through any additional parameters to the generation server, which can do validation of parameters.
+- Guardrails-related API updates would be additive, and we will not modify any OpenAI API elements, like `messages` or `choices`.
+    - We will add a `detectors` block in the request, containing `input` and `output` blocks. This will provide flexibility and control, allowing users to provide list of detectors with parameters for both chat completions input and output separately.
+    - Here, `input` and `output` does not strictly refer to whether only the `input` to chat completions is provided to the detectors or that only the `output` to chat completions is provided to the detectors. Detectors specified in `output` can potentially take both input and output of a chat completions model.
```
mdevino, line 16: I like this idea, but it's inconsistent with what we currently have for `/api/v1/task/classification-with-text-generation`.
Should we look into combining `/api/v1/task/classification-with-text-generation` and `/api/v2/text/generation-detection` in the future?

docs/architecture/adrs/005-chat-completion-support.md
```diff
@@ -0,0 +1,74 @@
...
+    - We will add a `detections` block in the response.
+- If no input or output detectors were requested, then the `detections.<missing element / input or output>` would not be present in the response.
+- If none of input or output detectors are requested, then we will return `422` (similar behavior to other standalone `v2` endpoints).
```
mdevino, line 19: From this I'm understanding all options below are invalid. Is this correct?

```json
{
    "detections": {}
}
```

```json
{
    "detections": {
        "input": {}
    }
}
```

```json
{
    "detections": {
        "output": {}
    }
}
```

```json
{
    "detections": {
        "input": {},
        "output": {}        
    }
}
```

docs/architecture/adrs/005-chat-completion-support.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Response behavior
+
+#### Assumptions regarding choices
+This is particular to the [`choices` array](https://platform.openai.com/docs/api-reference/chat/object#chat/object-choices) on the chat completion response.
+1. Each choice is independent.
+2. Each choice is generated with the same input.
+
+#### Response
+- If the LLM (model used to generate chat completions) doesn't generate `content` in any `message` in `choices`, then we will ignore the `choices` and return a warning about `detections.output`.
```
mdevino, line 30: I think we should only return the warning if `output` is present in the request.

mdevino: > perhaps it just needs to be explicitly written that output detectors are requested?
I think this is what I meant. In a scenario in which only `input` detectors are requested, we probably shouldn't send any warnings regarding `output`.
Does that make sense?

mdevino: I agree this would be implied. I was just unsure on how specific these ADRs need to be. :sweat_smile:

docs/architecture/adrs/005-chat-completion-support.md
gkumbhat: Yes, we will not be able to remove v1 API for quite a while.  We actually first need to create a `v2` replacement for text-generation and this chat completion would provide a solid framework for `v2/text-generation` when we get to it.

However, at this point, we need to be restricted with how things were working with `v1` endpoint.

Also, we do not plan to add such capability in `/generation` endpoint, since that will deviate its purpose.

docs/architecture/adrs/005-chat-completion-support.md
```diff
@@ -0,0 +1,74 @@
...
+
+### Response behavior
+
+#### Assumptions regarding choices
+This is particular to the [`choices` array](https://platform.openai.com/docs/api-reference/chat/object#chat/object-choices) on the chat completion response.
+1. Each choice is independent.
+2. Each choice is generated with the same input.
+
+#### Response
+- When output detectors are requested, if the LLM (model used to generate chat completions) doesn't generate `content` in any `message` in `choices`, then we will ignore the `choices` and return a warning about `detections.output`.
```
gkumbhat, line 30: this is probably until we have any detector that works on function / tool calling.

docs/architecture/adrs/005-chat-completion-support.md
```diff
@@ -0,0 +1,74 @@
...
+    - An alternate consideration made here was to not order the detections and just include `detector_id`. However, since detectors are called in parallel, just collecting results may lead to different ordering each call. This ordering will allow processing of detections with spans in order by `start`.
+
+
+##### Streaming response
+
+The streaming response with guardrails is built off the [chat completion chunk responses](https://platform.openai.com/docs/api-reference/chat/streaming). This applies when both streaming and output detectors are requested.
+- Each choice's `delta.content`, if present, will be the `content` used for `detections`. Like the unary case, if there is no `content` overall, a warning will be returned.
+- Since detector use will cause content to be "chunked" (e.g. collected into sentences for sentence detectors), to prevent ambiguity, the `delta.role` will be provided on each stream event, and `delta.content` will include the "chunked" text. This will apply to each `choice` in `choices`.
```
gkumbhat, line 40: Can we add a point here that this API will also follow the "only returning fully processed chunks" strategy, same as text generation ?

https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/main/docs/architecture/adrs/002-streaming-response-aggregation.md#:~:text=Only%20fully%20processed%20frames%20will

docs/architecture/adrs/005-chat-completion-support.md
```diff
@@ -0,0 +1,74 @@
...
+- For output detectors that run on the contents of an entire output stream, results will be included as `detections` on the second-last event (last one before `[DONE]`) of the stream, regardless of presence of `usage`.
+    - An alternate consideration made here was the currently implemented aggregation strategy, where if the user requested any output detectors that required the entire output, the stream events with detection results would only be returned all together. The problem here was the results would then not really be "streamed" back.
+
+
+#### User interactions
+
+This part gives a brief overview of how we would expect a user to access information from the response object.
+- Identifying which choice of potentially multiple choices that results exist for: `detections.output[0].choice_index` (will give index of choice in original `choices`)
+- Result info of a given choice: `detections.output[0].results[1].detection_type`
```
gkumbhat, line 50:
```suggestion
- Result info of a given choice: `detections.output[0].results[0].detection_type`
```

We should probably make it `results[0]` otherwise, it may look like we are starting index for results with `1`  to the reader

docs/architecture/adrs/005-chat-completion-support.md
```diff
@@ -0,0 +1,74 @@
...
+
+The streaming response with guardrails is built off the [chat completion chunk responses](https://platform.openai.com/docs/api-reference/chat/streaming). This applies when both streaming and output detectors are requested.
+- Each choice's `delta.content`, if present, will be the `content` used for `detections`. Like the unary case, if there is no `content` overall, a warning will be returned.
+- Since detector use will cause content to be "chunked" (e.g. collected into sentences for sentence detectors), to prevent ambiguity, the `delta.role` will be provided on each stream event, and `delta.content` will include the "chunked" text. This will apply to each `choice` in `choices`.
+- A `detections` block will be added to each stream event. Each processed `choice` will be noted with the inclusion of the choice's index in the original `choices` array by a `choice_index` field, even if there are no particular `results` from running the output detectors.
+- For output detectors that run on the contents of an entire output stream, results will be included as `detections` on the second-last event (last one before `[DONE]`) of the stream, regardless of presence of `usage`.
```
gkumbhat, line 42: I think it would be good to add a note here that we are not changing how `usage` gets returned or how any optional  response object will get return

docs/architecture/adrs/005-chat-completion-support.md
```diff
@@ -0,0 +1,74 @@
...
+- Results from input detectors will include `message_index` to indicate which message was processed. In cases where only the last message of the chat history may be processed to limit resource usage, this will help make it obvious to the user which message was processed.
+- Similarly, since output detectors will be run on all `choices` provided by the chat completions model, `choice_index` will be provided to inform the user which `choice` that detector result(s) apply for. This will apply regardless if the detector itself may take full or partial input to chat completions model.
+- A new result aggregation strategy will need to be implemented to put any detector results that require the full output text from a chat completions model (e.g. all the text of a particular `choice`) on the second-last event.
+- For the streaming response, implementation-wise, aggregating the streaming responses will be complicated since currently for the openAI chat completions API, each stream event returns content for only one choice, e.g.
```
gkumbhat, line 62: Shall we also add a point that each of the detector type will implement a "filtering" mechanism to select the messages they can work with ? 🤔

gkumbhat: yep, that sounds more accurate.

gkumbhat (approved):
Looks good to me! Thanks Evaline for the PR

Issue #195, opened 2024-09-13
Update standalone chat detection API in alignment with any chat API with generation


evaline-ju:
## Description

As an orchestrator developer, if I want to run any standalone detections with chat history, I want to make sure this standalone detection endpoint on chat is in line with any chat API that includes both generation and detection (e.g. role vs. author).

## Discussion

Provide detailed discussion here

## Acceptance Criteria

- [x] [Orchestrator API](https://foundation-model-stack.github.io/fms-guardrails-orchestrator/?urls.primaryName=Orchestrator+API) `/detection/chat` is updated in line with #167

PR #211, opened 2024-09-26
:wrench: Update standalone chat detections API

evaline-ju:
Related to updates for detections on chat completions and ADR proposed in #209 

This enables detections on chat messages in a standalone detections orchestrator endpoint, reusing the openAI schema for messages. This endpoint will focus on allowing for the concept of detectors running on entire message histories (e.g. multiple roles and content).

The corresponding chat detector endpoint has been updated as well.

For: #195

### DIFF
--- a/docs/api/openapi_detector_api.yaml
+++ b/docs/api/openapi_detector_api.yaml
@@ -146,20 +146,9 @@ paths:
                   Response Chat Analysis Unary Handler Api V1 Text Context Chat
                   Post
                 example:
-                  - message_index: 0
-                    results:
-                      - {
-                          "detection_type": "detection_type",
-                          "detection": "detection",
-                          "score": 0.999
-                        }
-                  - message_index: 1
-                    results:
-                      - {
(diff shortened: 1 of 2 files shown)


docs/api/openapi_detector_api.yaml
```diff
@@ -229,23 +226,16 @@ components:
...
-              author: system
-            - content: Hi, is this powered by siri or alexa?
-              author: user
-            - content: Better, it's watsonx
-              author: assistant
-            - content: This is awesome!
-              author: user
+          items:
+            allOf:
+              - $ref: https://raw.githubusercontent.com/openai/openai-openapi/master/openapi.yaml#/components/schemas/ChatCompletionRequestMessage
```
gkumbhat, line 234: hmm. this would include `name` etc as well. Just want to make a note of that. I don't think there is any harm in that.

gkumbhat (approved):
Looks good to me! Just added a note

PR #212, opened 2024-09-27
:wrench: Add detector_params as optional for every detectors endpoint

evaline-ju:
Previously only the passthrough was documented for one type of detector


gkumbhat (approved):
Looks good to me!

Issue #213, opened 2024-09-27
Startup client-dependent health checks should not be true by default


evaline-ju:
## Description

As discussed for #163 , orchestrator health / startup / liveliness / existence should not be gated by client status.

Anything client dependent should not be true by default: https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/main/src/main.rs#L44, as used here: https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/main/src/orchestrator.rs#L79

While users can toggle by env var, true by default means each user that integrates the newer version will have to be specifically aware.

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

PR #216, opened 2024-09-30
🔧 Turnoff default health check

gkumbhat:
Supports: #213 

### Changes

- Turn off default health startup check
- Fix `tonic-build` `compile_protos` deprecation issue by updating the package and changing `compile` to `compile_protos`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,9 +77,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.88"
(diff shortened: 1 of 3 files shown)


pscoro (approved):
LGTM, I wasn't sure whether we wanted it to run by default, but if we did want this to be true by default we would need to fix it anyway (ie `--no-start-up-health-check` or another fix, iirc `--start-up-health-check=false` was not parsing to false), this is simpler

PR #220, opened 2024-10-01
Client refactor

### LINKED ISSUE #142: Change detector calls, so that only hostname is needed in detectors configuration
## Description

As an orchestrator contributor, I want to change the code calling detector endpoints, so that I can set only the hostname in the `hostname` parameter in detectors configuration (e.g. `https://my-detector.com` instead of `https://my-detector.com/api/v1/text/generation`).

Note: This should be done for all detector endpoints.

## Discussion

This was raised as an issue as [a comment in one of the recent PRs](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/pull/116#discussion_r1697668762). Making this change would allow two things:

1. Support detectors that implement more than one endpoint ([detectors API reference](https://foundation-model-stack.github.io/fms-guardrails-orchestrator/?urls.primaryName=Detector+API)).
2. 

I've tried the following:

```diff
    /// Invokes detectors implemented with the `/api/v1/text/generation` endpoint
    pub async fn generation_detection(
        &self,
        model_id: &str,
        request: GenerationDetectionRequest,
    ) -> Result<Vec<DetectionResult>, Error> {
        let client = self.client(model_id)?;
-        let url = client.base_url().as_str();
+       let url = client.base_url().join("/api/v1/text/generation").unwrap();
        let response = client
-            .post(url)
+            .post(url.as_str())
            .header(DETECTOR_ID_HEADER_NAME, model_id)
            .json(&request)
            .send()
            .await?;
        if response.status() == StatusCode::OK {
            Ok(response.json().await?)
        } else {
            let error = response.json::<DetectorError>().await.unwrap();
            Err(error.into())
        }
    }
```

As a result, detector hostnames could be configured as desired. I also noticed that, if a detector was configured with a full url (say `https://mydetector.com/a/full/url`), the path was entirely replaced (i.e. `https://mydetector.com/a/full/url` becomes `https://mydetector.com/api/v1/text/generation`, not `https://mydetector.com/a/full/url/api/v1/text/generation`.

## Acceptance Criteria

### LINKED ISSUE #165: Add detector_type to configure different detector endpoints

### LINKED ISSUE #191: Add concept of provider type in orchestrator
## Description

Currently we only support 2 providers which can be chosen to provide "text generation". However, as we are designing chat completion use-cases, we want to enable another LLM generation provider like vLLM. We would want the ability to run multiple types of providers together, but since we don't want to take responsibility of routing, we want to restrict configuration of only 1 provider per provider type..

This story is to add the concept of provider type in orchestrator that can be used in `config.yaml` in orchestrator.

### Tasks
1. Create ADR for `provider_type` parameter
2. Implement `provider_type` concept in orchestrator
3. Add provider type to existing text generation clients, i.e TGIS and NLP
4. Add provider type support in `config.yaml`
5. Add provider type as `text_generation` in config for TGIS / NLP.

## Acceptance Criteria

- [ ] ADR merged to repo
- [ ] Add provider type for text-generation and NLP client
- [ ] Add provider type in config.yaml for already supported providers.

### LINKED ISSUE #194: Add openAI chat completions as a provider for chat [vllm support]
## Description

As an orchestrator developer, I want to be able to support chat generation models, so that I can eventually provide detections on them. 

Since [openAI chat completions is a popular API](https://platform.openai.com/docs/guides/chat-completions/getting-started) for chat and supported through [vllm today](https://github.com/vllm-project/vllm), by adding openAI chat API support as an additional "generation [server]" provider, orchestrator users will be able to use chat models and eventually be able to run detections on chat input/output.

## Discussion

`provider_type` could depend on #191 but this issue does not have to be gated on a final naming (as long as it is distinguishable from the current tgis/nlp provders]

This will start the implementation of the `v2/chat/completions-detection` without (or ignoring) the `detectors` section. Ideally openAI or vllm-related parameters will just be passed through (one strategy is via dynamic HashMap [silly small example](https://github.com/evaline-ju/fms-guardrails-orchestrator/tree/test-passthrough) but others welcome) for the chat generation provider.

## Acceptance Criteria

- [ ] READMEs are updated

### LINKED ISSUE #198: Add openAI completions as a provider for generation
## Description

As a orchestrator user, I want to be able to use [openAI completions](https://platform.openai.com/docs/api-reference/completions) as a generation provider, so that I can use generation servers that are serving the REST openAI (completions) API.

May be related to #191 in the sense of orchestrator can only expect one configured "text generation" provider at a time (whether tgis or nlp/caikit-nlp and with this addition here, openAI completions). Note: `/chat/completions` will be tracked separately via chat issues.

## Discussion

Current generation providers use GRPC [for tgis and caikit-nlp today]. OpenAI will bring in a REST API

## Acceptance Criteria

- [ ] READMEs are updated

### LINKED ISSUE #217: Refactor clients to better align with requirements
## Description
Several changes to refactor clients to better align with current usage patterns and new requirements.

## Tasks
- Update client implementations to wrap a single client rather than a map of multiple clients
  - Update `create_http_clients()` and `create_grpc_clients()` to `create_http_client()` and `create_grpc_client()` to create single clients
- Add `clients.detector` module with implementations for each detector
  - Drop existing `DetectorClient`
- Add `clients.openai` module with `OpenAiClient` implementation
- Add `GenerationProvider::OpenAi` variant, `ChatGenerationConfig`, `chat_generation` field to `OrchestratorConfig` (resolves #191)
  - Add config validation rule to only allow `GenerationProvider::OpenAi` provider for `ChatGenerationConfig`
- Add `DetectorType` to map detector client implementation (resolves #165) **
- Add `Client` trait including `health()` method, implement for all clients
  - Drop `HealthCheck` and `HealthProbe` traits
- ~~Add `Handler` trait to implement request handling for each client~~
- Add `ClientMap` type to hold all client types, update `Context`, replacing `generation_client`, `chunker_client`, and `detector_client` fields with `clients: ClientMap`
- Update `create_clients()`
- Update `Orchestrator::client_health()`
- Add config validation rule to ensure that only hostnames are set as opposed to urls (resolves #142)
- Inject http/grpc clients into `Client::new()` rather than creating in

declark1:
This PR is a consolidation of several changes related to clients.

- Refactors clients, including splitting detector implementations into standalone clients; see #217 for additional details.
  - `TextContextChatDetectorClient` is a placeholder and not fully implemented yet
- Adds `OpenAiClient` with chat completions and completions support
- Adds `GenerationProvider::OpenAi` variant
- Adds `ChatGenerationConfig` (`OrchestratorConfig.chat_generation` field)
- Adds `DetectorType` (`DetectorConfig.type` field)
- Requires that only the hostname (as opposed to url) is set for client services (`ServiceConfig.hostname` field) and adds hostname validation
- Health check related tweaks / simplifications
  - The `/info` response structure is changed as clients are no longer grouped by type. Example of new format with all changes:
      ```json
      {
	      "services": {
		      "generation": {
			      "status": "HEALTHY"
		      },
		      "chunker1": {
			      "status": "HEALTHY"
		      },
		      "detector1": {
			      "status": "UNKNOWN",
			      "code": 404,
			      "reason": "Not Found"
		      },
		      "detector2": {
			      "status": "UNKNOWN",
			      "code": 404,
			      "reason": "Not Found"
		      }
	      }
      }
      ```
- Moved `clients::Error` and `HttpClient` and related items out of `clients.rs` to separate files

Closes #217 
Closes #165 
Closes #142 
Closes #191 
Closes #194 
Closes #198


pscoro (approved):
Overall LGTM. I had one idea for a potential improvement that I was digging into on Friday but I think it may be over-engineered, I'll mention it anyway:

Every client implementor seems to have exactly one "inner client" thats either a tonic or http client. The tonic grpc clients need to be cloned in every handler function, and the http ones are accessed by reference. We could potentially add an associated `Inner` type and a `client()` to access it in the `Client` trait. For clients with an HTTP inner client, this is implemented to return `&HttpClient` and for tonic a clone of the client, e.g. `NlpServiceClient<LoadBalancedChannel>`. This on its own doesn't sound too bad but the catch is that it will introduce the need for `AnyClient`, because the ClientMap:
```rust
pub struct ClientMap(HashMap<String, Box<dyn Client>>);
```
will now need the client to specify the Inner type (e.g. `Client<Inner = &HttpClient>`), so instead we can use something like:
```rust
pub struct ClientMap(HashMap<String, Box<dyn AnyClient>>);
```
Setting up this type-erased blanket-implemented trait is doable and there may be more use cases for it in the future but for now I think it may just add more confusing code to the project without adding much value, all it would really add as benefit for now is that any client just needs to call `client()` to get their inner client, and in the case of grpc this means we could get rid of the clone line thats at the top of every handler.

src/clients/openai.rs
```diff
@@ -0,0 +1,423 @@
...
+    /// while lower values like 0.2 will make it more focused and deterministic.
+    #[serde(default)]
+    pub temperature: Option<f32>,
+    /// An alternative to sampling with temperature, called nucleus sampling,
+    /// where the model considers the results of the tokens with top_p probability mass.
+    /// So 0.1 means only the tokens comprising the top 10% probability mass are considered.
+    #[serde(default)]
+    pub top_p: Option<f32>,
+
+    // Additional vllm params
```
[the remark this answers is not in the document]
declark1: The reason it is included in this PR as the additional "provider" (issue references `ProviderType`, but took a different approach) for OpenAI is part of this PR, so it made since to include it. Is there any benefit of taking it out as it is already implemented (will create more work at this point)?

We have to maintain request/response types for all clients, so I don't really see an alternative to this. The API is pretty stable at this point as well.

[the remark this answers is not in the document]
gkumbhat: +1 on that.

We need to move towards a "passthrough" mechanism for this. Since we don't want to take API validation duty for vLLM or any other chat completion server.

Additionally, things change fast, there are already parameters deprecating and switching between vllm and openai and if we do strict validation, then we would end up being in the middle of their transition making our lives more difficult by having to constantly monitor and adhere to that transition.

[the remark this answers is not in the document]
declark1: This code isn't doing any validation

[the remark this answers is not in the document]
declark1: Per discussion, will update to support all modalities and add `max_completion_tokens` parameter so that it is fully compatible with the current spec.

src/clients/generation.rs
```diff
@@ -253,3 +234,19 @@ impl GenerationClient {
...
 }
+
+#[cfg_attr(test, faux::methods)]
+#[async_trait]
+impl Client for GenerationClient {
+    fn name(&self) -> &str {
+        "generation"
+    }
+
+    async fn health(&self) -> HealthCheckResult {
```
gkumbhat, line 245: Q: if there is a client (generation for example) that exposes http API, but implements health endpoint with a different API, how would we support that?

Just wanted to understand what would be the path for that kind of requirement, in case it happens.

src/config.rs
```diff
@@ -221,25 +258,81 @@ impl OrchestratorConfig {
...
                 }
             }
-            Ok(())
         }
+
+        // Generation config is valid
+        if let Some(generation) = &self.generation {
+            // Provider is valid
+            if !matches!(
+                generation.provider,
```
gkumbhat, line 304: Q: if we separate out `chat_provider` from `generation_provider`, then it would simplify this sort of validation right?

Any reasons for putting these providers together in same `enum` ?

src/health.rs
```diff
@@ -1,218 +1,128 @@
...
-pub trait HealthProbe {
-    /// Makes a health check request to each client and returns a map of client service ids to health check results.
-    fn health(
-        &self,
-    ) -> impl std::future::Future<Output = Result<HashMap<String, HealthCheckResult>, Error>> + Send;
-}
-
 /// Health status determined for or returned by a client service.
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "UPPERCASE")]
```
gkumbhat, line 13: nice!

src/orchestrator.rs
```diff
@@ -92,37 +111,34 @@ impl Orchestrator {
...
         info!("Performing start-up actions for orchestrator...");
         if health_check {
-            info!("Probing health status of configured clients...");
-            // Run probe, update cache
-            let res = self.clients_health(true).await.unwrap_or_else(|e| {
-                // Panic for unexpected behaviour as there are currently no errors propagated to here.
-                panic!("Unexpected error during client health probing: {}", e);
-            });
+            info!("Probing client health...");
+            let client_health = self.client_health(true).await;
```
gkumbhat, line 115: should this be donw in a tokio task so that it doesn't block startup ?  (we can address it in future, but just wanted to make a note here)

src/orchestrator.rs
```diff
@@ -162,50 +178,98 @@ fn get_chunker_ids(
...
+            GenerationProvider::OpenAi => unimplemented!(),
+        }
+    }
+
+    // Create chat generation client
+    if let Some(chat_generation) = &config.chat_generation {
+        match chat_generation.provider {
+            GenerationProvider::OpenAi => {
+                let client =
+                    create_http_client(DEFAULT_OPENAI_PORT, &chat_generation.service).await;
```
gkumbhat, line 221: Q: Is it possible to push the detail that this client needs `http_client` to its own module? 

so that we can just do `client = GenerationProvider:OpenAI::get_client(DEFAULT_OPENAI_PORT, &chat_generation.service)` 

that way the creation of client from this file for all, looks same?

gkumbhat (commented):
Nice refactor Dan.  Left a few comments, suggestions and questions

docs/api/orchestrator_openapi_0_1_0.yaml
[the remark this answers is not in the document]
declark1: Ah yea, it looks like my linter (OpenAPI extension) made these changes when I applied the health updates. I can disable it and revert if you prefer.

docs/architecture/adrs/006-detector-type.md
```diff
@@ -0,0 +1,45 @@
...
+
+The guardrails orchestrator interfaces with different types of detectors. 
+Detectors of a given are type are compatible with only a subset of orchestrator endpoints.
+In order to reduce changes of misconfiguration, we need a way to map detectors to be used only with compatible endpoints.
+
+
+## Decision
+
+We decided to add the `type` parameter to the detectors configuration. 
+Possible values are `text_contents`, `text_context_chat`, `text_generation` and `text_context_doc`.
```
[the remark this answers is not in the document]
declark1: Will push a separate commit to update the client name accordingly.

src/config.rs
declark1: Yes, I did this as I anticipate we will adopt the OpenAI API in the future instead of continuing to maintain the TGIS compatibility layer.

src/orchestrator.rs
declark1: I assume `Orchestrator::on_start_up()` tasks, i.e. checks, should "complete" prior to startup, but I don't really have an opinion on this. @pscoro any thoughts on this?

src/config.rs
declark1: I just pushed a change to drop `provider` from `ChatGenerationConfig` (since it is always openai for this) and `GenerationProvider::OpenAi` variant. Also dropped provider validation rules as they aren't needed now.

src/orchestrator.rs
gkumbhat: On one side, its good to check if everything is wired up correctly, but then in case connections and this verification takes time, it will halt orchestrator bootup and look like orchestrator is having issue.

So I think we should make this async and put out `error` levels logs if this startup check fails for some downstreams.

(can be handled in separate PR though)

src/clients/detector/text_context_doc.rs
```diff
@@ -0,0 +1,112 @@
...
+                .json::<DetectorError>()
+                .await
+                .unwrap_or(DetectorError {
+                    code,
+                    message: "".into(),
+                });
+            Err(error.into())
+        }
+    }
+}
```
mdevino, line 57: This method is quite similar in all detector clients (it basically invokes the detector and parse the response). The differences are the URL and the request/response types.

Should we create a default implementation for it using generics? Or maybe use some kind of template method? Or even create a helper function to abstract away repetitive code?

Note: definitely not something that needs to be done in this PR, I just wanted to start the discussion.

src/clients/http.rs
```diff
@@ -0,0 +1,139 @@
...
+                            error!(
+                                "unexpected http health check status code: {}",
+                                response.status()
+                            );
+                            HealthStatus::Unknown
+                        },
+                        code: response.status(),
+                        reason: response.status().canonical_reason().map(|v| v.to_string()),
+                    }
+                }
```
mdevino, line 79: Should we move this to a separate (probably inline) function to reduce nesting, or is this common in Rust?

src/config.rs
```diff
@@ -221,25 +254,66 @@ impl OrchestratorConfig {
...
+
+        // Chat generation config is valid
+        if let Some(chat_generation) = &self.chat_generation {
+            // Hostname is valid
+            if !is_valid_hostname(&chat_generation.service.hostname) {
+                return Err(Error::InvalidHostname(
+                    "`chat_generation` has an invalid hostname".into(),
+                ));
+            }
+        }
```
mdevino, line 314: Should we break these into separate functions instead of leaving comments (e.g. `validate_detector_config()`, `validate_chunker_config()` and `validate_generation_config()`?

mdevino (approved):
Changes look good to me!
Having detectors in separate files make it much easier to navigate the code.

I left a few comments around breaking down large functions to start a dicussion, but I'd suggest leaving these for a separate PR (assuming we agree on these changes).

src/orchestrator.rs
declark1: I do recall that we wanted to leverage health checks as an extension of config validation, so that we should catch misconfigured services at start-up.

src/orchestrator.rs
declark1: This change to inner client dependency injection, rather than creating the client in `<client_type>::new()`, was intentional for testing purposes.

I'm not sure I fully understand your suggestion though, so let's discuss during standup.

src/clients/detector/text_context_doc.rs
declark1: I actually initially created a helper function used by each implementation to reduce this boilerplate, but then I took it out in favor of keeping things explicit and I didn't want to assume that all detector implementations will have the exact same error format, header requirements, etc. I can add it back if you guys would prefer that.

src/clients/http.rs
declark1: I thought about trying to further simplify/reduce this logic, but it wasn't added as part of this PR, so I tried to keep it as close as possible. I did change/simplify how we build the `reason` values though.  I don't think the nesting is an issue here as its only 2 levels of `if/else`.

src/config.rs
declark1: Sure. Just pushed a change to split the rules into methods.

mdevino: Looks great. Thanks!

src/clients/detector/text_context_doc.rs
mdevino: I don't really have a strong opinion on this one. @evaline-ju, @gkumbhat, any thoughts?

src/clients/generation.rs
declark1: This `health()` method is very flexible, it just needs to return `HealthCheckResult`. It can be implemented to call any API.

A couple of current scenarios:

1. `NlpClient` has a separate `health_client` for calling the gRPC health service. For this, the `health()` method calls `self.health_client.check()` to perform the health checks.
2. The detectors were recently updated to move the health endpoint to a separate non-mTLS server listening on a different port. As such, they will need their own routes as well. This will require a configuration update and creating a second `health_client` (just like `NlpClient`) for the detector clients. The `health()` method will need to be updated to call `self.health_client.health()`. (I believe the plan is to do this in a separate PR after this is merged @mdevino?)

Let me know if this doesn't answer the question.

src/orchestrator.rs
gkumbhat: yep, thats accurate, but then we realized that this has a unwanted side-effect that it will make orchestrator's startup to fail when even 1 of the downstream is not responding.  And so it looks like orchestrator is having issue (from looking at deployment), but the real problem is one of the detectors. 

So the idea pivoted to orchestrator doing this check, but instead of blocking its own startup / boot, it will expose this information with API call.

src/clients/detector/text_contents.rs
```diff
@@ -0,0 +1,147 @@
...
+};
+
+#[cfg_attr(test, faux::create)]
+#[derive(Clone)]
+pub struct TextContentsDetectorClient {
+    client: HttpClient,
+}
+
+#[cfg_attr(test, faux::methods)]
+impl TextContentsDetectorClient {
```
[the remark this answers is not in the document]
declark1: I think any pre/post-processing should be handled outside of the client

src/orchestrator.rs
pscoro: > I assume `Orchestrator::on_start_up()` tasks, i.e. checks, should "complete" prior to startup, but I don't really have an opinion on this. @pscoro any thoughts on this?

This brings up a good point, the naming of that function may have some legacy from when we used to want orchestrator to fatally error if the start up health check failed. Now since we decided that the orchestrator does not treat its clients as dependencies.

I think `on_startup()` should imply behaviour executed before start up, maybe in the future this could include some tasks that need to run synchronously. I think what we should do is run the call to `client_health` in a new tokio task thats started inside the `on_start_up()`. Thoughts on this?

src/orchestrator.rs
gkumbhat: What I meant is, each of the client module, like `src/clients/detector/text_chat.rs` or `nlp.rs`, knows if they can work with grpc or http.

So am thinking, the previous paradigm that created client inside these module was nice from `orchestratrator.rs` perspective, since it just needed to initialize the nlp client and it would call the `create_grpc_client` function etc. 

With this change, this information that a client is gRPC or http, now lives in 2 places.

So my question was, if we can move invocation of `create_grpc_client` (or `create_clients`)  inside the same module as the client and expose just a more consistent `get_client` function over here.

So we need not call this `crate_grpc_client` or `create_clients` function from inside the constructor like before, but just add that as a static method or something.


So what if it was something like:
```rust
impl NlpClient {
    pub fn new(...) {...}  
    pub fn create_client(...) -> Self {
     // call to create_grpc_client
   }
```

and this would potentially solve the testing problem you were seeing and at the same time keep the client type (grpc vs http) information inside the module 


thoughts?

src/orchestrator.rs
pscoro: > I do recall that we wanted to leverage health checks as an extension of config validation, so that we catch misconfigured services at start-up. I believe this was the original intention of running probes at start-up.

This would still catch misconfig errors at start up time. Since we are already decided that the state of clients (as they are known via their configs) does not effect the liveliness of the orchestrator, and are only reported as non-fatal errors, I dont think there is functionally any difference between:
- starting an async health check, then starting the orchestrator & recving requests, then handling and reporting the response of the health check without crashing on failure, and 
- starting a sync health check, waiting for the response, then handling and reporting the response of the health check without crashing on failure, and then starting the orchestrator

src/orchestrator.rs
pscoro:
```
TgisClient::new(client);
```
The way I see it, a client and its inner client are highly coupled, so i think this is a good point to bring up, since the creation process requires making the inner client in a separate step.

On friday I was trying to make the inner client part of the client trait itself:
```rust
#[async_trait]
pub trait Client: Send + Sync + 'static {
    type Inner;

    /// Returns the inner client.
    fn client(&self) -> Self::Inner;
    ...
```
This way it could all conform to a single trait. This runs into type erasure complexity because we are aggregating a hashmap of client impls with different `Inner` client types, we would need an `AnyClient` to use in place of `Client<Inner = ...>`. Even though going down this route is probably too much for now, I still think the client and its inner client sound highly coupled so we could hide the inner service as @gkumbhat suggests.

src/clients/detector/text_context_doc.rs
gkumbhat: Actually being explicit might work out for us better, given the variability of detectors and even implementation of detectors. This would provide us a way to handle differences more clearly at cost of some duplication, but I think thats okay.

src/orchestrator.rs
declark1: Just to clarify, the rationale for moving inner client construction out of `<client>::new()` is to easily enable injecting mock http/grpc clients which are not created through `create_http_client()` and `create_grpc_client()`. Decoupling this provides greater flexibility for testing purposes.

declark1: > injecting mock http/grpc clients which are not created through create_http_client() and create_grpc_client()

Thinking about this further, I don't think this will be necessary. I'll revert this back and we can adjust if needed once we figure out our mocking strategy.

gkumbhat (approved):
Looks good to me! Thanks for all the updates and aligning clients design with goals.

PR #222, opened 2024-10-03
:wrench: Remove context from chat endpoint in detector API

evaline-ju:
From #211 and before, there hasn't been the concept of `context` for the chat endpoint. Removing the reference to avoid confusion


gkumbhat (approved):
Looks good to me!

PR #224, opened 2024-10-08
Client optional health port configuration

pscoro:
This PR is for resolving https://github.ibm.com/ai-foundation/watson-fm-stack-tracker/issues/1373, which is resulting from detectors now using 8001 for health check while still using 8000 for the main port.

This PR enables health ports to be optionally configured as part of the `service` of a client in the orchestrator config. If omitted, the health port will default to the value of the main port for that client (which itself may default). Alongside adding this, the PR does a minor refactor of how the service configs are passed down, since there is currently complexity in the orchestrator for client creation that could have been caught earlier as part of config loading/validation. 

(Note that as a follow up to this, we can also be resolving client tls cert paths before client creation, but that is outside the scope of this PR.)

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -10,7 +10,8 @@ generation:
     provider: tgis # tgis or nlp
     service:
         hostname: localhost
-        port: 8033
+        port: 8033 # Optionally, the default value of the port (dependent on the client type) can be overridden
+        health_port: 8035 # Optionally, the using `port` for the health endpoint can be overridden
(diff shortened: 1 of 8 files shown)


pscoro:
Just realized there was still a piece that I'm missing so I updated the PR to draft for the meantime

pscoro:
Closing due to sufficient fix provided part of https://github.com/foundation-model-stack/fms-guardrails-orchestrator/pull/220, we can open follow up issues if we want to make further configuration improvements

Issue #228, opened 2024-10-14
whole doc chunker not working with text_contents detectors


evaline-ju:
## Describe the bug

Post #220, use of `whole_doc_chunker` for use with `text_contents` type detectors is resulting in errors

## Expected behavior

`whole_doc_chunker` usable for `/text/detection/content` standalone endpoint and any generation endpoints with classification 

## Observed behavior

500 for the end-user, with `unwrap` on the (nonexistent) chunker client causing errors at https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/main/src/orchestrator/unary.rs#L766

PR #229, opened 2024-10-14
:bug: Fix whole_doc_chunker missing from client list

gkumbhat:
closes: #228 

### Changes

- Add `whole_doc_chunker` client as default client for chunkers

### DIFF
--- a/src/clients.rs
+++ b/src/clients.rs
@@ -41,7 +41,6 @@ pub mod http;
 pub use http::HttpClient;
 
 pub mod chunker;
-pub use chunker::ChunkerClient;
 
 pub mod detector;
 pub use detector::TextContentsDetectorClient;
--- a/src/clients/chunker.rs
+++ b/src/clients/chunker.rs
@@ -39,7 +39,7 @@ use crate::{
 const DEFAULT_PORT: u16 = 8085;
 const MODEL_ID_HEADER_NAME: &str = "mm-model-id";
 /// Default chunker that returns span for entire text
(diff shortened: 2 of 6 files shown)


src/clients.rs
```diff
@@ -41,7 +41,6 @@ pub mod http;
 pub use http::HttpClient;
 
 pub mod chunker;
-pub use chunker::ChunkerClient;
```
gkumbhat, line 44: Since this is unused, clippy was complaining

mdevino (approved):
Looks good to me (though I'm not super familiar with the chunking piece yet).
I've tested changes locally and verified that the whole_doc_chunk works for the two endpoints below:

* /api/v1/task/classification-with-text-generation
* /api/v1/task/server-streaming-classification-with-text-generation

declark1 (approved):
LGTM,

Just a couple of nit suggestions:
- I'd rename `chunker::DEFAULT_MODEL_ID` to `DEFAULT_CHUNKER_ID` instead of import aliases
- I'd move `tokenize_whole_doc` and `tokenize_whole_doc_stream` out of `chunker.rs` to `unary.rs` and `streaming.rs`, respectively, as this functionality is no longer part of the client.

gkumbhat:
Thanks Dan. I renamed the chunker model id to `DEFAULT_CHUNKER_ID`.. For moving the functions out, I am not sure yet, since while they are not part of the client, they are providing similar functionality that this client provide 🤔  

We can discuss and address that later.

Issue #210, opened 2024-09-26
Implement standalone chat detection API for text


evaline-ju:
## Description

As a followup to #195, as an orchestrator developer, I want to implement the changes for the updated API, so that I can provide standalone detections on chat. This endpoint is designed for detectors working on full chat histories

### Tasks
1. Implement standalone detector endpoint `/text/detection/chat` (accepting chat history request) in orchestrator
2. Add http client for `detection_chat`
3. Test with dummy detector that works on full chat histories

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

mdevino:
The [API](https://foundation-model-stack.github.io/fms-guardrails-orchestrator/?urls.primaryName=Detector+API#/default/chat_analysis_unary_handler_api_v1_text_chat_post) wasn't very clear o a few things to me, so I'm looking [here](https://platform.openai.com/docs/api-reference/chat).

1. Do we need to support deprecated fields (Assistant message's `function_call`, Function message)?

![image](https://github.com/user-attachments/assets/acb23c1c-4011-4de0-8b42-bd5bee6e7976)


2. Will we support all modalities for this endpoint? e.g.: For User messages' `content` array variant, will we support both text and image?

![image](https://github.com/user-attachments/assets/89d3bbd4-b56f-4bf5-b0a8-20cd760ab3d2)

PR #233, opened 2024-10-17
Chat standalone endpoint

mdevino:
This addresses #210. Messages can be either strings or text content types.

Tests performed:
1. Accepts string messages
2. Accepts text content type messages
3. Rejects image content type messages
4. Propagates 404 messages from detectors.
5. Propagates 503 messages from detectors.
6. Propagates 422 messages from detectors.
7. Returns error 500 when detectors return messages not compliant with [the API](https://foundation-model-stack.github.io/fms-guardrails-orchestrator/?urls.primaryName=Detector+API#/default/chat_analysis_unary_handler_api_v1_text_chat_post).

### DIFF
--- a/src/clients/detector/text_chat.rs
+++ b/src/clients/detector/text_chat.rs
@@ -1,10 +1,13 @@
 use async_trait::async_trait;
+use hyper::{HeaderMap, StatusCode};
+use serde::Serialize;
 
-use super::DEFAULT_PORT;
+use super::{DetectorError, DEFAULT_PORT, DETECTOR_ID_HEADER_NAME};
 use crate::{
-    clients::{create_http_client, Client, HttpClient},
+    clients::{create_http_client, openai::Message, Client, Error, HttpClient},
     config::ServiceConfig,
     health::HealthCheckResult,
+    models::DetectionResult,
 };
 
 #[cfg_attr(test, faux::create)]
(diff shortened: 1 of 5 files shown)


src/clients/detector/text_chat.rs
```diff
@@ -29,9 +32,37 @@ impl TextChatDetectorClient {
...
-    pub async fn text_chat(&self) {
-        let _url = self.client.base_url().join("/api/v1/text/chat").unwrap();
-        todo!()
+    pub async fn text_chat(
+        &self,
+        model_id: &str,
+        request: ChatDetectionRequest,
+        headers: HeaderMap,
+    ) -> Result<Vec<DetectionResult>, Error> {
+        let url = self.client.base_url().join("/api/v1/text/chat").unwrap();
```
gkumbhat, line 41: nit: Can we move this string `/api...`  to constants ?

src/models.rs
```diff
@@ -939,6 +943,76 @@ pub struct ContextDocsResult {
...
+        }
+        Ok(())
+    }
+
+    /// Validates if content type array contains only text messages
+    fn validate_content_type(&self, content: &Content) -> Result<(), ValidationError> {
+        match content {
+            Content::Array(content) => {
+                for content_part in content {
+                    if content_part.r#type != "text" {
```
gkumbhat, line 996: Is there a way to restrict this based on the full endpoint they are using ? 

Since we would only want to restrict this for `/text/detection/chat`  endpoints and not for `image`, for example

src/orchestrator/unary.rs
```diff
@@ -447,6 +450,61 @@ impl Orchestrator {
...
+            .collect::<Vec<_>>();
+
+            Ok(ChatDetectionResult { detections })
+        });
+        match task_handle.await {
+            // Task completed successfully
+            Ok(Ok(result)) => Ok(result),
+            // Task failed, return error propagated from child task that failed
+            Ok(Err(error)) => {
+                error!(request_id = ?task.request_id, %error, "detection on chat content task failed");
```
gkumbhat, line 497: nit:
```suggestion
                error!(request_id = ?task.request_id, %error, "detection task on chat failed");
```

src/orchestrator/unary.rs
```diff
@@ -447,6 +450,61 @@ impl Orchestrator {
...
+                Err(error)
+            }
+            // Task cancelled or panicked
+            Err(error) => {
+                let error = error.into();
+                error!(request_id = ?task.request_id, %error, "detection on chat content task failed");
```
gkumbhat, line 503: nit:
```suggestion
                error!(request_id = ?task.request_id, %error, "detection task on chat failed");
```

gkumbhat (commented):
small nits, otherwise, looks good. Thanks for adding this

src/models.rs
```diff
@@ -939,6 +943,98 @@ pub struct ContextDocsResult {
...
+        for (model_id, detector_params) in &self.detectors {
+            if let Some(stream) = detector_params.get("stream") {
+                match stream.as_bool() {
+                    Some(stream) => {
+                        if stream {
+                            return Err(ValidationError::Invalid(format!("`stream` parameter specified for model `{model_id}` must not be true")));
+                        } else {
+                            return Ok(());
+                        }
+                    }
```
gkumbhat, line 990: I think we don't need the validation of `stream` here. Since it sort of assumes what detector supports or not. 

If the detector doesn't support any parameter, it would be on the detector implementation to validate it.

mdevino: @gkumbhat I've removed this validation.

gkumbhat (approved):
Looks good to me!

Issue #227, opened 2024-10-14
Pass runtime detector_params for contents detectors


evaline-ju:
## Description

As a follow-up to #212, and as a orchestrator user, I want to be able to pass runtime parameters through `detector_params` for detectors implementing the `/contents` endpoint like for the `/context/doc` endpoint, so that I can provide runtime parameters on calls.

## Discussion

This will affect both the standalone orchestrator endpoint for `/contents` and generation endpoints that utilize this.

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

Issue #234, opened 2024-10-22
threshold parameter may cause detector validation failures


evaline-ju:
## Describe the bug

`threshold` is currently a parameter only used by the orchestrator to filter scores but is passed as "detector parameters" along with other detector-specific parameters to the detectors.

Some detectors will validate that there are no extra parameters passed. Since `threshold` is used by the orchestrator, this parameter should not be passed on to detectors.

## Sample Code

```
curl -k 'POST' \
  'https://hostname/api/v2/text/detection/chat' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
    "messages": [
        {
          "content": "Tell me a bad joke",
          "role": "user"
        }
    ],
    "detectors": {
      "detector-id": {"threshold": 0.5, "temperature": 0.0}
    }
}'
```

## Expected behavior

`threshold` is only used for result filtering

## Observed behavior

Potential failure on extra inputs to detector

PR #235, opened 2024-10-22
:sparkles::bug: Pass on non-threshold detector parameters

evaline-ju:
- Add `detector_params` to additional detectors API endpoints (`/text/contents` and `/text/generation`) as documented through #212 
- As discovered in #234, we only use `threshold` in the orchestrator, so we do not want to pass this on to the detectors. This is currently just removed. This now builds off #236 which assumes `pop_threshold` will only be done once per detector flow
- Unit tests update show the threshold parameter no longer being passed on

Closes: #227 , #234

### DIFF
--- a/src/orchestrator/streaming.rs
+++ b/src/orchestrator/streaming.rs
@@ -36,7 +36,7 @@ use crate::{
     models::{
         ClassifiedGeneratedTextStreamResult, DetectorParams, GuardrailsTextGenerationParameters,
         InputWarning, InputWarningReason, TextGenTokenClassificationResults,
-        TokenClassificationResult, THRESHOLD_PARAM,
+        TokenClassificationResult,
     },
     orchestrator::{
         unary::{input_detection_task, tokenize},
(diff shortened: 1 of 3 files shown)


src/orchestrator/streaming.rs
```diff
@@ -277,6 +278,7 @@ async fn streaming_output_detection_task(
         // Get the default threshold to use if threshold is not provided by the user
         let default_threshold = detector_config.default_threshold;
         let threshold = detector_params.threshold().unwrap_or(default_threshold);
+        detector_params.remove("threshold");
```
mdevino, line 281: Since you've setup a constant for this parameter, should we use the constant here instead*?

src/orchestrator/streaming.rs
```diff
@@ -352,9 +355,11 @@ async fn generation_broadcast_task(
 /// Consumes chunk broadcast stream, sends unary requests to a detector service,
 /// and sends chunk + responses to detection stream.
 #[instrument(skip_all)]
+#[allow(clippy::too_many_arguments)]
```
mdevino, line 358: This doesn't have to be done now, but isn't this a sign that we should rethink our structs?

src/clients/detector/text_chat.rs
```diff
@@ -109,6 +109,7 @@ pub struct ChatDetectionRequest {
     /// Chat messages to run detection on
     pub messages: Vec<Message>,
 
+    /// Detector parameters
```
mdevino, line 112:
```suggestion
    /// Detector parameters (available parameters depend on the detector)
```
In case this makes sense, could you also update it in other classes?

mdevino (changes requested):
Looks good, overall. Left a few questions.

src/orchestrator/streaming.rs
mdevino: I created #237, so we don't forget about this.

gkumbhat (approved):
Looks good to me!

PR #240, opened 2024-10-30
Implement Chat Completions API

### LINKED ISSUE #192: Implement chat completion API in orchestrator [no detections]
## Description

Add chat completion endpoint in orchestrator that allows calling out to chat completion client (with no detection)

In this story, we are assuming that chat completion client will be added in separate story.

## Acceptance Criteria

- [ ] Chat completion endpoint added in orchestrator
- [ ] Streaming error events are handled correctly

declark1:
This PR implements the OpenAI Chat Completions Detections API (no detections).

- Conditionally enables the chat completions detection endpoint (`/api/v2/chat/completions-detection`) if chat generation config is provided
- Adds `ChatCompletionsDetectionTask` and `Orchestrator::handle_chat_completions_detection` handler that calls `OpenAiClient::chat_completions()` (handles both unary and streaming)

As `orchestrator/unary.rs` is getting _very_ large, starting with this, I've created a standalone file `orchestrator/chat_completions_detection.rs` for the handler. It also feels weird to add this to "unary" as its both a unary and streaming endpoint. I suggest that we consider opening a PR to split the other handlers into their own files as we have added many new endpoints.

Closes #192

### DIFF
--- a/src/clients/openai.rs
+++ b/src/clients/openai.rs
@@ -58,9 +58,9 @@ impl OpenAiClient {
     #[instrument(skip_all, fields(request.model))]
     pub async fn chat_completions(
         &self,
-        request: ChatCompletionRequest,
+        request: ChatCompletionsRequest,
         headers: HeaderMap,
-    ) -> Result<ChatCompletionResponse, Error> {
+    ) -> Result<ChatCompletionsResponse, Error> {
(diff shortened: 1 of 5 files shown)


declark1:
> I assume this has been tested to make sure any of the variety of chat completions parameters are successfully passed through?

I tested several of them but not all, just because there are so many.

Issue #238, opened 2024-10-23
Provide more info for unwraps when they error


resoluteCoder:
## Description

As a user, I want to be able to increase the information I get back when things go wrong so that I can be able to locate the source of the error more easily. 

## Discussion

This is more of an initiative to handle the various `unwrap` functions throughout the code base. 

When I encountered the `whole_doc_chunker` bug I received `task 51 panicked with message "called Option::unwrap() on a None value"`. This error message doesn't exactly point to which `unwrap` panicked and why. 

I would say to implement some combination of `match` expressions, or at the very least sub `unwrap` for `expect` in some cases. For example as for the one above `expect("Failed to get ChunkerClient/chunker id")` 

I do see there are several `unwrap_or_else` functions in the code base that could be used as well

I know this wouldn't apply to the `unwrap` functions that are used in those instances where there couldn't be an error but that's another discussion :smile: 

I'm willing to take on this initiative.

gkumbhat:
Thanks for opening the issue @resoluteCoder .. Can you propose a few changes like that in a PR ?

resoluteCoder:
@gkumbhat Absolutely, sounds good! :smile:

resoluteCoder:
@gkumbhat I created #241 as a proposed solution with several ideas. Please let me know your thoughts. :smile: 

While going through the codebase I found several (11?) statements such as this. Referring to the `get_as` function
```rust
let client = ctx.clients.get_as::<TextContentsDetectorClient>(&detector_id).unwrap();
```

I see in here that we could potentially handle the errors there instead of propagating the error up / having the calling function handle it. 
```rust
pub fn get_as<V: Client>(&self, key: &str) -> Option<&V> {
    self.0.get(key)?.downcast_ref::<V>()
}
```
I say this because we could reduce the repetition of handling the same error in multiple places as well as eliminate the ambiguity of the cause of the error.

To explain the error ambiguity further. Correct me if I'm wrong but if the `?` propagates the error up or the downcast ref returns `None` both would be handled by the calling unwrap not knowing which one actually failed to return a value.

I would propose that the possible errors are handled separately in that function and only return a concrete value otherwise do the panic in there. 

I hope this makes sense and please feel free to point out any issues that you have. 

Thanks for the long read :smile:

PR #241, opened 2024-10-31
changing unwraps to expect or unwrap_or_else

resoluteCoder:
This PR attempts to start adding error messages for the `unwrap` functions as follows:

- Used `unwrap_or_err` for functions that return results with an err.
- Used expect for functions that return an option.
- Used the convention of `error getting x b/c of y` since it seems thats a pattern of the repo

Since this is a draft I'd like to get feedback on the specificity of error messages. Such as having the error message say `not being to accessing a value from provided key` or just generically `not getting x value`.

Once I know what pattern(s) that need to be implemented I can tackle more but this should spark some conversations.

I also kept with the panics  as a first step and would love to handle the errors more appropriately later on.

Addresses #238

### DIFF
--- a/src/clients.rs
+++ b/src/clients.rs
@@ -199,8 +199,11 @@ pub async fn create_http_client(default_port: u16, service_config: &ServiceConfi
         Some(_) => "https",
         None => "http",
     };
-    let mut base_url = Url::parse(&format!("{}://{}", protocol, &service_config.hostname)).unwrap();
-    base_url.set_port(Some(port)).unwrap();
+    let mut base_url = Url::parse(&format!("{}://{}", protocol, &service_config.hostname))
+        .unwrap_or_else(|e| panic!("error parsing base url: {}", e));
+    base_url
+        .set_port(Some(port))
+        .unwrap_or_else(|_| panic!("error setting port: {} ", port));
     let request_timeout = Duration::from_secs(
         service_config
             .request_timeout
--- a/src/orchestrator/streaming.rs
+++ b/src/orchestrator/streaming.rs
@@ -270,11 +270,18 @@ async fn streaming_output_detection_task(
         // Create a mutable copy of the parameters, so that we can modify it based on processing
         let mut detector_params = detector_params.clone();
         let detector_id = detector_id.to_string();
-        let chunker_id = ctx.config.get_chunker_id(&detector_id).unwrap();
+        let chunker_id = ctx
+            .config
+            .get_chunker_id(&detector_id)
+            .expect("chunker id is not found");
 
         // Get the detector config
         // TODO: Add error handling
-        let detector_config = ctx.config.detectors.get(&detector_id).unwrap();
+        let detector_config = ctx
+            .config
+            .detectors
+            .get(&detector_id)
+            .expect("detector config not found");
 
         // Get the default threshold to use if threshold is not provided by the user
         let default_threshold = detector_config.default_threshold;
@@ -285,7 +292,7 @@ async fn streaming_output_detection_task(
         // Subscribe to chunk broadcast stream
         let chunk_rx = chunk_broadcast_streams
             .get(&chunker_id)
-            .unwrap()
+            .expect("chunker receiver not found")
             .subscribe();
         let error_tx = error_tx.clone();
         tokio::spawn(detection_task(
@@ -391,7 +398,7 @@ async fn detection_task(
                             let client = ctx
                                 .clients
                                 .get_as::<TextContentsDetectorClient>(&detector_id)
-                                .unwrap();
+                                .expect("text contents detector client not found");
                             match client.text_contents(&detector_id, request, headers)
                                 .await
                                 .map_err(|error| Error::DetectorRequestFailed { id: detector_id.clone(), error }) {
--- a/src/orchestrator/streaming.rs
+++ b/src/orchestrator/streaming.rs
@@ -291,6 +291,7 @@ async fn streaming_output_detection_task(
         tokio::spawn(detection_task(
             ctx.clone(),
             detector_id.clone(),
+            detector_params,
             threshold,
             detector_tx,
             chunk_rx,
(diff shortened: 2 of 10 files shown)


src/clients.rs
```diff
@@ -204,9 +204,15 @@ pub async fn create_http_client(default_port: u16, service_config: &ServiceConfi
...
     };
-    let mut base_url = Url::parse(&format!("{}://{}", protocol, &service_config.hostname)).unwrap();
-    base_url.set_port(Some(port)).unwrap();
+
+    let mut base_url = Url::parse(&format!("{}://{}", protocol, &service_config.hostname))
+        .unwrap_or_else(|e| panic!("error parsing base url: {}", e));
+    base_url
+        .set_port(Some(port))
+        .unwrap_or_else(|_| panic!("error setting port: {} ", port));
+
```
declark1, line 213: I just want to note that we simply `unwrap()` here as urls are validated prior to this during the config validation step. However, it doesn't hurt to add good panic messages here just in case.

e.g.
https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/main/src/config.rs#L305

src/orchestrator/streaming.rs
```diff
@@ -271,11 +271,17 @@ async fn streaming_output_detection_task(
...
+            detector_id
+        ));
 
         // Get the detector config
         // TODO: Add error handling
-        let detector_config = ctx.config.detectors.get(&detector_id).unwrap();
+        let detector_config = ctx.config.detectors.get(&detector_id).expect(&format!(
+            "error getting detector config with detector: {}",
+            detector_id
+        ));
```
declark1, line 284: If a client requests an invalid `detector_id`, we should return an error to the user rather than panic and terminate the task. I can't recall if we already check this before processing, but the comment above leads me to believe that we don't. Need to review further.

src/orchestrator/unary.rs
```diff
@@ -270,13 +270,21 @@ impl Orchestrator {
                         let ctx = ctx.clone();
                         let detector_id = detector_id.clone();
                         let detector_params = detector_params.clone();
-                        let detector_config = ctx.config.detectors.get(&detector_id).unwrap();
+                        let detector_config = ctx.config.detectors.get(&detector_id).expect(
+                            &format!("error retrieiving config value for key {}", detector_id),
```
declark1, line 274: This has a typo "retrieiving", but I think we should simplify to something like "detector `<detector_id>` not found". Related to prior comment, we should check this and return an error to the user rather than panic.

src/orchestrator/unary.rs
```diff
@@ -754,7 +762,10 @@ pub async fn detect_for_generation(
     let client = ctx
         .clients
         .get_as::<TextGenerationDetectorClient>(&detector_id)
-        .unwrap();
```
declark1, line 757: FYI - these `clients.get_as::<T>()` calls should be "safe" unwraps as long as the `detector_id` is valid, which we should check prior to this

declark1 (commented):
@resoluteCoder thanks for looking into this -- we can definitely improve error handling in several places. I left a few comments inline.

A couple of notes:
- Some things are checked during config validation, e.g. urls, chunker_id for detectors, etc, but that isn't clear as there aren't comments indicating this. It definitely doesn't hurt to add explicit panic messages anyway though.
- Based on some changes you have here, it looks like we need to add validation for requested `detector_id`(s) prior to processing the requests and return an error to the user rather than panicing (need to review further to confirm we don't already do this, this would definitely be an important gap to close)

pscoro (approved):
LGTM, these improvements to the panic messages thrown by these unwraps seems reasonable, I left one nit about one of the lines. 

I think ideally we can handle some of the logic in the places where these unwraps exist during config validation to avoid needing to even have unwraps here (e.g. already resolve the base url or client certs paths), although IIRC making this improvement would require some less-trivial changes to the config deserialization

src/orchestrator/streaming/aggregator.rs
```diff
@@ -142,7 +142,9 @@ impl ResultActor {
         result.token_classification_results.output = Some(detections);
         if input_start_index == 0 {
             // Get input_token_count and seed from first generation message
-            let first = generations.first().unwrap();
+            let first = generations.first().expect(
+                "error retrieving first element in classified generated text stream result",
+            );
```
declark1, line 147: I think this is ok for now, but there should always be a first generation by the time we get to this point.

Initial thoughts for a potential future enhancement:

Above, we request generations from the `GenerationActor`:
```rust
// Get subset of generation responses relevant for this chunk
let generations = self
    .generation_actor
    .get_range(input_start_index, input_end_index)
    .await;
```
Currently, `get_range()` can panic if the start or end index is out of range. We should probably return `Option` and retry if `Option::None` or something until we have the generations needed.

src/orchestrator/unary.rs
```diff
@@ -845,7 +856,7 @@ pub async fn detect_for_context(
     let client = ctx
         .clients
         .get_as::<TextContextDocDetectorClient>(&detector_id)
-        .unwrap();
+        .expect("error getting text context doc detector client");
```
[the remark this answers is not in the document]
declark1: I didn't really comment on the messages, but I'm thinking we should probably standardize around "x not found" instead of "error getting x", etc. throughout the code.

In this case, we do validate that a requested `detector_id` exists prior to getting to this point, so I think that is why we currently `unwrap()` here and in some other places.

resoluteCoder:
@declark1 @pscoro 

Thanks for the comments so to summarize what I'm seeing is that some of the values are being validated elsewhere in the code but having `expect` or `unwraps_or_else` with messages is still a good idea.

Also standardize the error messages to simply say `x is not found` for those messages

If I'm understanding this correctly I can update this PR and remove draft for final review.

declark1:
> @declark1 @pscoro
> 
> Thanks for the comments so to summarize what I'm seeing is that some of the values are being validated elsewhere in the code but having `expect` or `unwraps_or_else` with messages is still a good idea.
> 
> Also standardize the error messages to simply say `x is not found` for those messages
> 
> If I'm understanding this correctly I can update this PR and remove draft for final review.

Yep, sounds good

resoluteCoder:
> Yep, sounds good. Also be sure to sign your commits to pass DCO.

Updated and I believe I was able to verify both commits now. Hopefully that will resolve the DCO error

resoluteCoder:
@declark1 Hmmm doesn't look like the DCO error is going away even though I see `Verified` for both of the commits. 

Is there any way to rerun or refresh the DCO action?

resoluteCoder:
@evaline-ju Noted. I followed the advice of the DCO and I think it broke things. :sweat_smile: 

I may just create another PR with reference to this one and close. What a mess, my apologies.

PR #242, opened 2024-10-31
[DRAFT] HTTP and gRPC client middleware for metrics collection

pscoro:
This PR introduces middleware for both HTTP and gRPC clients allowing metrics to be collected implicitly without overhead in instrumented functions (e.g. client handlers).

Some details still need to be finalized but this should build/function. Leaving in draft until finished revisions but opening for review since this PR introduces significant changes that I think should get looked at sooner rather than later

## Remaining TODO items
- some clean up and code deduplication
- some potential tweaks to allow more "custom" metrics to be collected for given endpoints

## Changes
- introduces HTTP client middleware using `reqwest_middleware` (popular well-maintained thin wrapper for `reqwest` that allows middlewar layers to be added, unfortunately requires modifying usages of `reqwest::Client` to `reqwest_middleware::Client`, but this is essentially just `reqwest::Client` + middleware layers
- there is no good tonic middleware crates. Creating our own middleware solution or maintaining a tonic middleware crate is likely overkill, opted instead to modify grpc clients implementation, adding a `GrpcClient` and `grpc_call!`, `grpc_stream_call!` macros to handle the metrics. The macros were a necessary step up in abstract to forego limitations in Rust's type system regarding Futures and lifetimes
- All handler functions decluttered by these changes

### DIFF
--- a/src/clients/chunker.rs
+++ b/src/clients/chunker.rs
@@ -15,31 +15,30 @@
 
 */
 
-use std::pin::Pin;
-
 use async_trait::async_trait;
 use axum::http::HeaderMap;
-use futures::{Future, Stream, StreamExt, TryStreamExt};
+use futures::{Stream, StreamExt, TryStreamExt};
 use ginepro::LoadBalancedChannel;
-use tonic::{Code, Request, Response, Status, Streaming};
-use tracing::{info, instrument};
+use hyper::StatusCode;
+use tonic::{Request, Response, Status, Streaming};
+use tracing::instrument;
 
 use super::{
-    create_grpc_client, errors::grpc_to_http_code, grpc_request_with_headers, BoxStream, Client,
-    Error,
+    create_grpc_client, grpc::GrpcClient, grpc_request_with_headers, BoxStream, Client, Error,
 };
+use crate::pb::grpc::health::v1::HealthCheckResponse;
 use crate::{
     config::ServiceConfig,
+    grpc_call, grpc_stream_call,
     health::{HealthCheckResult, HealthStatus},
     pb::{
         caikit::runtime::chunkers::{
-            chunkers_service_client::ChunkersServiceClient,
+            chunkers_service_client::ChunkersServiceClient, chunkers_service_server,
             BidiStreamingChunkerTokenizationTaskRequest, ChunkerTokenizationTaskRequest,
         },
         caikit_data_model::nlp::{ChunkerTokenizationStreamResult, Token, TokenizationResults},
-        grpc::health::v1::{health_client::HealthClient, HealthCheckRequest},
+        grpc::health::v1::{health_client::HealthClient, health_server, HealthCheckRequest},
     },
-    tracing_utils::trace_context_from_grpc_response,
 };
 
 const DEFAULT_PORT: u16 = 8085;
@@ -53,15 +52,29 @@ type StreamingTokenizationResult =
 #[cfg_attr(test, faux::create)]
 #[derive(Clone)]
 pub struct ChunkerClient {
-    client: ChunkersServiceClient<LoadBalancedChannel>,
-    health_client: HealthClient<LoadBalancedChannel>,
+    client: GrpcClient<ChunkersServiceClient<LoadBalancedChannel>>,
+    health_client: GrpcClient<HealthClient<LoadBalancedChannel>>,
 }
 
 #[cfg_attr(test, faux::methods)]
 impl ChunkerClient {
     pub async fn new(config: &ServiceConfig) -> Self {
-        let client = create_grpc_client(DEFAULT_PORT, config, ChunkersServiceClient::new).await;
-        let health_client = create_grpc_client(DEFAULT_PORT, config, HealthClient::new).await;
+        let client = create_grpc_client(
+            chunkers_service_server::SERVICE_NAME,
+            DEFAULT_PORT,
+            config,
+            ChunkersServiceClient::new,
+            true,
+        )
+        .await;
+        let health_client = create_grpc_client(
+            health_server::SERVICE_NAME,
+            DEFAULT_PORT,
+            config,
+            HealthClient::new,
+            false,
+        )
+        .await;
         Self {
             client,
             health_client,
@@ -74,12 +87,12 @@ impl ChunkerClient {
         model_id: &str,
         request: ChunkerTokenizationTaskRequest,
     ) -> Result<TokenizationResults, Error> {
-        let mut client = self.client.clone();
         let request = request_with_headers(request, model_id);
-        info!(?request, "sending client request");
-        let response = client.chunker_tokenization_task_predict(request).await?;
-        trace_context_from_grpc_response(&response);
-        Ok(response.into_inner())
+        grpc_call!(
+            self.client,
+            request,
+            ChunkersServiceClient::chunker_tokenization_task_predict
+        )
     }
 
     #[instrument(skip_all, fields(model_id))]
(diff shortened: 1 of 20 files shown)


src/clients/chunker.rs
```diff
@@ -53,15 +52,29 @@ type StreamingTokenizationResult =
...
+    client: GrpcClient<ChunkersServiceClient<LoadBalancedChannel>>,
+    health_client: GrpcClient<HealthClient<LoadBalancedChannel>>,
 }
 
 #[cfg_attr(test, faux::methods)]
 impl ChunkerClient {
     pub async fn new(config: &ServiceConfig) -> Self {
-        let client = create_grpc_client(DEFAULT_PORT, config, ChunkersServiceClient::new).await;
-        let health_client = create_grpc_client(DEFAULT_PORT, config, HealthClient::new).await;
+        let client = create_grpc_client(
```
pscoro, line 62: Explanation:

because of prost autogenerated protobuf, information that would be runtime string information in HTTP (eg url passed to client), is static type information in grpc, making tracing certain details as string fields added to spans difficult, since rust does not support dynamic type reflection. I still dont have a great solution for getting a given grpc request's service method/endpoint in the middleware, but we can at least hold on to the service name with this method.

Also: the new boolean as the last arg to `create_grpc_client` is for enabling tracing. The thought process being that health clients typically dont want to have tracing enabled (causes middleware to skip collecting metrics on every request, allows health requests to run faster/frequently)

src/clients/chunker.rs
```diff
@@ -108,28 +125,32 @@ impl Client for ChunkerClient {
...
+            grpc_request_with_headers(HealthCheckRequest { service: "".into() }, HeaderMap::new());
+        let response: Result<HealthCheckResponse, Error> =
+            async { grpc_call!(self.health_client, request, HealthClient::check) }.await;
         let code = match response {
-            Ok(_) => Code::Ok,
-            Err(status) if matches!(status.code(), Code::InvalidArgument | Code::NotFound) => {
-                Code::Ok
-            }
-            Err(status) => status.code(),
+            Ok(_) => StatusCode::OK,
```
pscoro, line 134: grpc_call caused things to need to be rewritten but the logic should be the same, will see if I can reduce/clean up these changes

src/clients/chunker.rs
```diff
@@ -108,28 +125,32 @@ impl Client for ChunkerClient {
...
-    async fn health(&self) -> HealthCheckResult {
-        let mut client = self.health_client.clone();
-        let response = client
-            .check(HealthCheckRequest { service: "".into() })
-            .await;
+    async fn health(&self) -> Result<HealthCheckResult, Error> {
+        let request =
+            grpc_request_with_headers(HealthCheckRequest { service: "".into() }, HeaderMap::new());
+        let response: Result<HealthCheckResponse, Error> =
+            async { grpc_call!(self.health_client, request, HealthClient::check) }.await;
```
pscoro, line 132: TODO: undecided if `let request = async { grpc_call!(self.health_client, request, HealthClient::check) }.await;` is too messy or if i should modify the macro slightly to avoid needing the async block when not immediately returning the value of the `grpc_call!` in a handler function

src/clients/chunker.rs
```diff
@@ -108,28 +125,32 @@ impl Client for ChunkerClient {
         "chunker"
     }
 
-    async fn health(&self) -> HealthCheckResult {
-        let mut client = self.health_client.clone();
-        let response = client
-            .check(HealthCheckRequest { service: "".into() })
-            .await;
+    async fn health(&self) -> Result<HealthCheckResult, Error> {
```
pscoro, line 128: The return type of health endpoints did not previously have any cases where errors could happen and so the return type was not wrapped in `Result<>` until now. This caused problems with code in the new grpc macro expecting a result type. It could be worked around I think, by modifying the macro, but I was also thinking for consistency all handlers should `-> Result<...>`, where for health checks the errors are still turned into appropriate health check responses. I am undecided on this detail and whether its worth looking for a change in the macro to avoid needing to make this change

src/clients/errors.rs
```diff
@@ -69,6 +69,23 @@ impl From<reqwest::Error> for Error {
     }
 }
 
+impl From<reqwest_middleware::Error> for Error {
```
pscoro, line 72: Identical to handling of `reqwest::Error`, except for the thin middlware wrapper crate `reqwest_middleware`

src/clients/grpc.rs
```diff
@@ -0,0 +1,310 @@
...
+        Self {
+            service_name,
+            inner,
+            tracing_enabled,
+        }
+    }
+}
+
+/// Work around for compiler bug: https://github.com/rust-lang/rust/issues/96865
+pub trait SendFuture: Future {
```
pscoro, line 37: Avoids a compiler bug that seems to have existed for a couple years and has been well documented, regarding `Send` and Future dyn objects, that would otherwise occure on line 212 from:
```
let client_result: Pin<
            Box<
                dyn Future<Output = Result<tonic::Response<tonic::Streaming<_>>, tonic::Status>>
                    + Send,
            >,
        > = Box::pin($method(&mut client, $request));
```

src/clients/grpc.rs
```diff
@@ -0,0 +1,310 @@
...
+
+        let trace_id = current_trace_id();
+        let protocol = "gRPC";
+        let service_name = $client.service_name.clone();
+        let request_metadata = $request.metadata();
+
+        info!(?trace_id,
+            protocol,
+            service_name,
+            // method,
```
pscoro, line 72: still looking for a "clean" solution to getting the specific grpc method name in these metrics. For the most part, this information can already be found through traces (ie by looking at the instrumented function name that parent these metric info events, which would be the handler for the endpoint/method being accessed. But i imagine it would still be nice to have the specific type name of the grpc service method being called serialized here if possible

src/clients/http.rs
```diff
@@ -22,15 +22,17 @@ use url::Url;
...
 use crate::health::{HealthCheckResult, HealthStatus, OptionalHealthCheckResponseBody};
 
+pub type ClientImpl = reqwest_middleware::ClientWithMiddleware;
+
 #[derive(Clone)]
 pub struct HttpClient {
     base_url: Url,
     health_url: Url,
-    client: reqwest::Client,
+    client: ClientImpl,
```
pscoro, line 31: The switch from `reqwest::Client` to `reqwest_middleware::Client` makes essentially 0 logical changes and hopefully isnt too intrusive of a change

pscoro (commented):
Some explanatory comments

src/clients/openai.rs
```diff
@@ -126,11 +126,11 @@ impl Client for OpenAiClient {
         "openai"
     }
 
-    async fn health(&self) -> HealthCheckResult {
+    async fn health(&self) -> Result<HealthCheckResult, Error> {
```
pscoro, line 129: TODO: rebased some recent openai changes i think, just realized that broke my build, need to update something to resolve the conflict between these two "thin wrappers" (`reqwest_middleware` and `reqwest_eventsource`). Looking into this

declark1:
> there is no good tonic middleware crates

Doesn't everyone use tower middleware for tonic? e.g. https://github.com/hyperium/tonic/tree/master/examples/src/tower

pscoro:
This PR was just a draft to showcase experimentation - the resulting discussion lead to a change of direction from `reqwest` to `hyper`: https://github.com/foundation-model-stack/fms-guardrails-orchestrator/pull/252/files - closing this PR now

Issue #248, opened 2024-11-07
Migrate openai client and HTTP streaming support from `reqwest` to `hyper`


pscoro:
## Description
As a follow up to https://github.com/foundation-model-stack/fms-guardrails-orchestrator/issues/247 we need to also support HTTP streaming without using any `reqwest` related crates. The previously implementation using `reqwest_eventsource` will need to be replaced. The openai client is the only client to currently use HTTP streaming (rest are gRPC), so it may make sense to port the openai client from `reqwest` to `hyper` on its own as part of this issue.

## Acceptance Criteria
- sensible solution to HTTP streaming support using `hyper`
- openai client uses `hyper` and whatever replacement streaming implementation in order to source SSE events

gkumbhat:
closing as the PR is merged

Issue #247, opened 2024-11-07
Migrate clients implementation from `reqwest` to `hyper`


pscoro:
## Description
As part of efforts to introduce middleware sensibly into the orchestrator we would like to first migrate away from `reqwest` and instead use `hyper` for the underlying HTTP clients. We want to do this in order to avoid limitations and conflicts imposed by `reqwest` that switching to a lower level library would avoid. In particular, we care about using `tower::Service` for middleware in the near future.

## Acceptance Criteria
- all http clients use Request, RequestBuilder, Client, etc from `hyper` instead of `reqwest` (the openai client may be exempt as HTTP streaming will have its own complexities
- all handler functions are updated to build and send `hyper` requests
- `hyper` clients are properly configurable and tls certs are being handled without `reqwest` (likely using `hyper-rustls`)

gkumbhat:
Closing as PR is merged now

PR #252, opened 2024-11-12
Migrate clients implementation from `reqwest` to `hyper`

pscoro:
This PR addresses https://github.com/foundation-model-stack/fms-guardrails-orchestrator/issues/247 and https://github.com/foundation-model-stack/fms-guardrails-orchestrator/issues/248. Decoupling these two issues proved difficult:

- leaving just open ai temporarily using a `reqwest` client would require some temporary hacking
- integrating `hyper` without considering SSE streaming support could cause issues retroactively

So this PR addresses HTTP streaming as well through the implementation of an `EventSource` (similar to `eventsource` crate) that uses a `hyper` Client instead of a `reqwest` Client and RequestBuilder. This PR also introduces additional infrastructure needed for configuring TLS without help from the `reqwest` library. Parts of the new TLS logic and streaming logic inspired by `reqwest` and `reqwest_eventsource`. The request invocation logic for all the HTTP clients needed to be updated, however all the new complexity is hidden in the `HttpClient` and the handler functions of individual clients becomes simplified.


declark1:
Nice work on this, I recognize this isn't trivial. I did a high-level review and just want to provide some initial feedback & suggestions.

- I would avoid adding custom types and traits that aren't absolutely essential, especially those with conflicting names of common types used in these crates as this can be confusing. This adds bloat and makes it harder for others to follow.

    A few callouts:
    1. This `Body` trait seems probably unneeded and the name is confusing as it conflicts with `hyper::body::Body`
    ```rust
    pub trait Body:
        hyper::body::Body<Data = Bytes, Error = hyper::Error> + Serialize + Send + Sync
    {
    }
    ```
    2. The `DataStream` type appears to be essentially the same thing as `http_body_util::BodyDataStream`, so I think you can use this instead of adding a custom type + `Stream` impl on it.  `http_body_util::BodyExt` includes `into_data_stream()` to return a `BodyDataStream` from `Body`, e.g. in `Response::bytes_stream()`, I think you can do the following and drop this type, unless I'm missing something:
    ```rust
        pub fn bytes_stream(self) -> impl futures_core::Stream<Item = Result<Bytes, Error>> {
            // DataStream(self.0.into_body().map_err(Into::into))
            self.0.into_body().map_err(Into::into).into_data_stream()
        }
    ```
    
    3. Is a `HttpClientExt` trait needed to add an `inner()` method? Why not just add `inner()` to `HttpClient` directly?
    ```rust
    pub trait HttpClientExt: Client {
        fn inner(&self) -> &HttpClient;
    }
    ```
    
    4. We probably don't need this `RequestLike` trait; for `HttpClient::get()` and `HttpClient::post()`, we can probably just simplify the type signature from `T: RequestLike + Send + Sync + 'static` to `body: impl Serialize` (or `body: Bytes` if we need more flexibility, handling the json serialization in the clients)?
    ```rust
    pub trait RequestLike: Serialize + Clone {}
    impl<T> RequestLike for T where T: Serialize + Clone + Send + Sync + 'static {}
    ```

- Hmm, I'm not a fan of adding the `EventSource` wrapper to this crate, it feels like it should be separate from the orchestrator, if we actually need it. I was playing around with `eventsource_stream` a bit, and I think we can get away with just using this directly for our requirements, doing something like the following for example in the openai client, i.e. just using `eventsource()` provided by the `eventsource_stream::Eventsource` trait.

```rust
    let mut event_stream = self
        .client()
        .clone() // ideally not needed
        .post(url, headers, request)
        .await
        .unwrap() // since this is just an example
        .bytes_stream()
        .eventsource(); // provided by eventsource_stream::Eventsource
    
    tokio::spawn(async move {
        while let Some(Ok(event)) = event_stream.next().await {
            // process similar to how we are doing it now
        }
    });
```
(if you drop it, I would keep it for reference if we ever need something like it)

- Why do `HttpClient::get()` and `HttpClient::post()` take `self` instead of `&self`?

pscoro:
@declark1 thanks for the initial feedback, really helped clean up the branch. I think I have addressed all of the comments

src/clients/detector.rs
```diff
@@ -49,29 +47,60 @@ impl From<DetectorError> for Error {
...
+    async fn post_to_detector<U: DeserializeOwned>(
+        &self,
+        model_id: &str,
+        url: Url,
+        headers: HeaderMap,
+        request: impl Serialize,
+    ) -> Result<U, Error> {
+        let mut headers = headers;
+        headers.append(DETECTOR_ID_HEADER_NAME, model_id.parse().unwrap());
+        let response = self.inner().clone().post(url, headers, request).await?;
```
declark1, line 87: Drop unnecessary clone

src/clients/http.rs
```diff
@@ -15,41 +15,175 @@
...
+/// HTTP response type, thin wrapper for `hyper::http::Response` with extra functionality.
+pub struct Response(pub hyper::http::Response<BoxBody<Bytes, hyper::Error>>);
+
+impl Response {
+    /// Deserializes the response body as JSON into type `T`.
+    pub async fn json<T: DeserializeOwned>(self) -> Result<T, Error> {
+        let data = self
+            .0
+            .into_body()
+            .boxed()
```
declark1, line 48: Drop unnecessary `boxed()` (body is `BoxBody`)

src/clients/http.rs
```diff
@@ -15,41 +15,175 @@
...
-            client,
+            inner,
         }
     }
 
     pub fn base_url(&self) -> &Url {
         &self.base_url
     }
 
+    pub fn endpoint(&self, path: impl Into<&'static str>) -> Url {
```
declark1, line 102: Simplify `path` type to `path: &str`

src/clients/detector.rs
```diff
@@ -49,29 +47,60 @@ impl From<DetectorError> for Error {
...
+    async fn post_to_detector<U: DeserializeOwned>(
+        &self,
+        model_id: &str,
+        url: Url,
+        headers: HeaderMap,
+        request: impl Serialize,
+    ) -> Result<U, Error>;
+
+    /// Wraps call to inner HTTP client endpoint function.
+    fn endpoint(&self, path: impl Into<&'static str>) -> Url;
```
declark1, line 73: Simplify `path` type to `path: &str`

src/clients/http.rs
```diff
@@ -15,41 +15,175 @@
...
+            Some(headers_mut) => {
+                headers_mut.extend(headers.clone());
+                debug!(
+                    ?url,
+                    ?headers,
+                    // If the body can not be serialized it will be caught below, we don't want to
+                    // complicate logging by handling string serialization error here.
+                    body = serde_json::to_string(&body).unwrap_or("".to_string()),
+                    "sending client request"
+                );
```
declark1, line 149: Move `headers_mut.extend()` after `debug!()` and drop the unnecessary clone. 

Hmm, I don't really like the `serde_json::to_string()` embedded here. Since all of our types impl `Debug`, we can alternatively update `body: impl Serialize + std::fmt::Debug` throughout and simplify this to:
```rust
debug!(?url, ?headers, ?body, "sending client request");
```

src/orchestrator/errors.rs
```diff
@@ -38,6 +39,55 @@ pub enum Error {
...
 }
 
+impl Error {
+    /// Helper function that logs an error level trace event if the given client error is internal.
+    fn handle_internal_client_error(error: &clients::Error, from_client: String) {
+        debug!(?error, "error received from {}", from_client);
+        if matches!(error, clients::Error::Internal { .. }) {
+            error!("internal client error: {}", error);
+        }
+    }
```
declark1, line 49: All of these error helper functions feel unnecessary IMO and out of scope for this PR.

declark1 (commented):
Nice work! I left a few comments inline.

src/clients/http.rs
pscoro: > alternatively update body: impl Serialize + std::fmt::Debug throughout

Going back to the `RequestLike` empty trait from before: In this case what is your opinion on having a trait to blanket `Serialize + Debug` so that we can simplify it to `impl RequestLike` or `impl IntoRequest` or something instead of making every function signature lengthier (also just like the idea that if I ever need to know what contrains a request or need to make modifications to it, its a single source of truth rather than several function signature changes)

(also addressed other nits and will push changes momentarily)

declark1: Yea, I agree it makes it easier to have a single source of truth if we need to change it later.

src/orchestrator/errors.rs
pscoro: All other points addressed directly. For this one, I simplified to a singular function. The (now called) `handle_client_error` function exists so the orchestrator can transform `clients::Error` appropriately everywhere where it is returned, in particular:
- any errors related to some request are turned into HTTP errors to propagate to the request sender, where the internal error details are omitted
- any internal client errors are immediately logged when received by the orchestrator before error is turned into HTTP variant

declark1: @pscoro I'm still not a fan of this; IMO maintaining an additional enum for each client type just for error handling seems unnecessary. Thinking about this more, I have some ideas to improve+simplify our errors (and I'm sure you do too), e.g. I think we can drop the "<Client>RequestFailed" variants and have a single `Client` variant encapsulating sufficient detail/context, etc.

As error handling enhancements are beyond the scope of this PR, can we drop these changes and address in a separate PR after we discuss? It appears they aren't required for the hyper migration unless I'm missing something.

src/clients/http.rs
declark1: Changes on this look good, thanks!

src/orchestrator/errors.rs
pscoro: @declark1 fully understandable, I'll try to trim out all the error changes from this PR today and we can look into this further as a separate item. I kinda already figured the change I pushed yesterday wasn't ideal and would cause more discussion, the only requirements I wanted to preserve here in this PR is that error into client response logic is kept correct and that internal error messages (about impl details) dont end up in http error responses.

> IMO maintaining an additional enum for each client type just for error handling seems unnecessary. Thinking about this more, I have some ideas to improve+simplify our errors (and I'm sure you do too), e.g. I think we can drop the <Name>RequestFailed variants and have a single Client variant encapsulating sufficient detail/context, etc.

Yep this is the start of a larger discussion for sure (we should book time to discuss), fully agree that client type/kind shouldnt exist only for error handling, and whether it should exist is part of a larger design discussion. I didnt feel like this is an appropriate place to introduce it but just wanted to tie up the work I was doing yesterday and push an update even though I hadnt thought of a simpler alternative yet

src/orchestrator/unary.rs
```diff
@@ -65,82 +65,85 @@ impl Orchestrator {
...
-                    task.inputs.clone(),
-                    task.text_gen_parameters.clone(),
-                    headers.clone(),
-                )
-                .await?;
-                debug!(?generation_results);
-                // Do output detections
-                let output_detectors = task.guardrails_config.output_detectors();
-                let output_detections = match output_detectors {
+        let task_handle = tokio::spawn(
```
pscoro, line 68: Just noticed - this file contains a lot of automatic reformatting because adding `.instrument(Span::current())` to async blocks (that need explicit instrumentation) causes `tokio::spawn(async move {` to break the `async move {` to a new line, incrementing indentation, iirc no functional changes besides the instrumentation

declark1:
@pscoro I just realized that our `connect_timeout` and `request_timeout` aren't migrating correctly as hyper doesn't implement these and they are being mapped to different types of timeouts below in the builder.

```rust
    let client = hyper_util::client::legacy::Client::builder(TokioExecutor::new())
        .http2_keep_alive_timeout(connect_timeout)
        .pool_idle_timeout(request_timeout)
        .build(https_conn);
```

I'm mostly concerned with `request_timeout` as this is important to have. I think we would either need to handle this in `HttpClient::send()` doing something like the example [here](https://github.com/hyperium/hyper/issues/2448#issuecomment-788184793) or use/create a wrapper such as [hyper-timeout](https://github.com/hjr3/hyper-timeout) (haven't used it before).

pscoro:
@declark1 `hyper-timeout` has connect, read, and write requests, would you know if request timeouts map to read timeouts, or to both read and write timeouts, etc?

declark1:
> @declark1 `hyper-timeout` has connect, read, and write requests, would you know if request timeouts map to read timeouts, or to both read and write timeouts, etc?

I'm not sure, it needs to be investigated further

pscoro:
Pushed a fix @declark1, some notes:
- before we used `reqwest::ClientBuilder::timeout()` for request timeout which sets the total timeout for the transmission of the request from the start of the connection until the end of the response - this is more closely achieved with `tokio::timeout` than with read/write timeouts
- this fix uses `hyper-timeout` for connection timeout and uses tokio timeout for request timeout
- Im not a fan of propagating request timeout into the http client to be used in the send function but its the most straightforward solution and I am currently already looking into changes relating to client creation and figure this will probably get addressed better in the near future
- to build on that, I think a cleaner solution for request timeouts is to use `tower-timeout` https://docs.rs/tower-timeout/latest/tower_timeout/index.html, which I can introduce with the tower integration

declark1 (approved):
LGTM, thanks @pscoro! Let's hold off on merging until full tests are ran.

src/clients.rs
```diff
@@ -210,53 +215,36 @@ pub async fn create_http_client(default_port: u16, service_config: &ServiceConfi
         .set_port(Some(port))
         .unwrap_or_else(|_| panic!("error setting port: {}", port));
     debug!(%base_url, "creating HTTP client");
+
+    let connect_timeout = Duration::from_secs(DEFAULT_REQUEST_TIMEOUT_SEC);
```
gkumbhat, line 219: we are making connect timeout = request timeout here, is this intentional ?

gkumbhat (approved):
Looks good to me! Thanks @pscoro for contribution

Issue #253, opened 2024-11-12
Cancellation of generation requests


resoluteCoder:
## Describe the bug

When a request is canceled—either due to manual intervention or connectivity issues—while waiting for a response from one of the following endpoints: `/api/v1/task/classification-with-text-generation` or `/api/v1/task/server-streaming-classification-with-text-generation`, the generator continues processing the request, even though there is no valid route for the response to return to. 

Additionally, the orchestrator does not generate any warnings or errors in these instances.

## Platform

Orchestrator: Latest main branch
Generator: https://github.com/opendatahub-io/vllm-tgis-adapter

## Sample Code
Simplest way to reproduce would be to use a curl command against one the endpoints then `ctrl c`

```bash
curl -H "Content-Type: application/json" --data '{"model_id": "gpt2", "inputs": "dummy input"}' http://localhost:8085/api/v1/task/classification-with-text-generation
```

## Expected behavior

I would expect that when a request is canceled, the generation process should be stopped. This would help reduce the load on the generator by preventing wasted computations and also lower the costs associated with unnecessary generation.

## Observed behavior

What you see happening (error messages, stack traces, etc...)

- No error/warnings from the orchestrator when connection is terminated

In this example you can see 6 concurrent requests being processed even though those 6 requests have been terminated
```bash
INFO 11-12 08:45:29 metrics.py:449] Avg prompt throughput: 0.0 tokens/s, Avg generation throughput: 2.6 tokens/s, Running: 6 reqs, Swapped: 0 reqs, Pending: 0 reqs, GPU KV cache usage: 0.9%, CPU
 KV cache usage: 0.0%.    
```

I would be more than happy to take on this issue :smile:

declark1:
Thanks for opening this @resoluteCoder. Just curious, how do you cancel the request in vLLM? Is it cancelled automatically if the connection is closed?

resoluteCoder:
@declark1 i do believe so. I see this issue here where the first comment answered the same question. https://github.com/vllm-project/vllm/issues/8332

resoluteCoder:
@declark1 Just verified with a quick grpcurl against vllm that once I canceled the command it went from 2 request back down to 0. :+1: :smile:

resoluteCoder:
Closing issue due to merged PR #254

PR #254, opened 2024-11-14
added ability to cancel streaming generation when client disconnects

resoluteCoder:
This PR aims to partially solve #253 by implementing the ability to cancel generation requests for the streaming endpoint in the event a client disconnects.


declark1 (changes requested):
Thanks for this fix @resoluteCoder! A couple of comments:

- When matching a single pattern, use the `if let` syntax instead of `match`
- The `response_tx.send()` isn't being awaited
- I think you meant to match the error of `result`

Suggested:
```rust
if let Err(e) = result {
    debug!(%trace_id, "could not send to client: {e}");
    let _ = response_tx.send(Err(Error::Cancelled)).await;
    return;
}
```

declark1:
Disregard my initial comment

The original code passes result: Result<ClassifiedGeneratedTextStreamResult, Error> through to the response channel. I don't believe your change here does what is intended; it would just send Error::Cancelled to the response channel if there was an error sending the result to it (which is unlikely).

resoluteCoder:
@declark1 

For some clarification are you saying that this PR doesn't do what I'm thinking it does or are you saying that its just this that needs to be changed `let _ = response_tx.send(Err(Error::Cancelled));`

resoluteCoder:
The first round is showing that gen completes with this new addition (verifying I didnt break anything) :sweat_smile: 

The second round shows me cancelling a request from the client (grpcurl) and generation stopping

Left side is generation server
Top right is orchestrator
Bottom right is client


https://github.com/user-attachments/assets/60093a2b-1587-4486-9221-651eb7d51bef

resoluteCoder:
@declark1 My understanding of this, is its matching to see if sending the result to the channel will error. To my knowledge this could occur when a channel becomes closed such as a disconnect from the client. This is why I wrote this here and it seemingly works. I'm not sure if I'm grasping the entire context of the streaming method or not but these are my thoughts 
```rust
 let response = response_tx.send(result).await;
 if let Err(e) = response {...}
 ```
Changed to the if let syntax as well. Let me know if I am still not understanding things correctly, Thanks! :smile:

declark1:
>  To my knowledge this could occur when a channel becomes closed such as a disconnect from the client. This is why I wrote this here and it seemingly works.

This sounds correct. I believe this works because: 
1. Client terminates connection, response channel is closed
2. `response_tx.send(result).await` returns an error because the response channel is closed
3. You added a match and `return` for this error scenario, which closes the receiver end of the `generation_stream`, terminating the generation server connection

I want to test a couple of things with this today and will get back.

declark1:
Due to the reasons noted above, this does solve the issue for streaming generation when there are no output detectors applied. With output detectors, the streaming pipeline is significantly more complex and we need to pass cancellation signals through the pipeline and terminate accordingly.

We do have an error broadcast channel (`error_tx` / `error_rx`) for error notification & task cancellation signaling.

I will add review comments with suggested changes to solve for this part.

declark1 (changes requested):
Thanks @resoluteCoder. See requested change inline. Can you also make the following changes to solve for this when output detections are applied (or alternatively I can push a commit to your branch if you give me write access)? Can you also run your tests with these changes? Thanks!

1. Add `tracing::warn` import
2. Update the block below accordingly. This similarly captures and logs the response channel closed error and broadcasts a cancellation signal to the other tasks.

  ```rust
  // Forward generation results with detections to response channel
  tokio::spawn(async move {
      let mut error_rx = error_tx.subscribe();
      loop {
          tokio::select! {
              Ok(error) = error_rx.recv() => {
                  error!(%trace_id, %error, "task failed");
                  debug!(%trace_id, "sending error to client and terminating");
                  let _ = response_tx.send(Err(error)).await;
                  return;
              },
              result = result_rx.recv() => {
                  match result {
                      Some(result) => {
                          debug!(%trace_id, ?result, "sending result to client");
                          if (response_tx.send(result).await).is_err() {
                              warn!(%trace_id, "response channel closed (client disconnected), terminating task");
                              // Broadcast cancellation signal to tasks
                              let _ = error_tx.send(Error::Cancelled);
                              return;
                          }
                      },
                      None => {
                          info!(%trace_id, "task completed: stream closed");
                          break;
                      },
                  }
              }
          }
      }
  });
  ```

3. Add a log statement to the other tasks when they receive the cancellation signal sent above. There should be 3 of these in generation_broadcast_task, detection_task, and chunker_broadcast_task.

  Replace
  ```rust
  _ = error_rx.recv() => { break },
  ```
  With:
  ```rust
  _ = error_rx.recv() => {
      warn!("cancellation signal received, terminating task");
      break
  },
  ```
Now the logs should look like below
image

resoluteCoder:
@declark1 sounds good. Updated the code :smile:

declark1 (approved):
Thanks @resoluteCoder, this looks great now!
@gkumbhat approving this, but let us know if you want to have a look before merging.

gkumbhat:
Thanks @declark1 for reviewing the PR and @resoluteCoder for the contribution and making all the updates.

Issue #193, opened 2024-09-13
Add function in content type detector to implement rules


gkumbhat:
## Description

As a part of designing chat API, we realized that certain type of detectors can only be applied to certain parts of the input / output. For example,  we would not want to apply `content type` of detectors on all the input "messages", since that may be duplicate and extra processing that can make whole thing suboptimal and have waste computations.

To avoid this, we decided that each detector type (implemented in issue #165 ) will correspond to set of rules that it will invoke that will involve things like validations, selecting parts of the messages it can work on, and apply these rules in `request` before passing it on to the detectors itself.

In this issue, we will design how we will implement such rules and implement these for "content type" detectors as an example.

Note: At the time of writing, these rules are only envisioned to be applicable in the context of chat completion API.

## Acceptance Criteria

- [ ] ADR around implementation and design of such rules merged
- [ ] Rules added to content type detector and PR merged

PR #255, opened 2024-11-14
:memo: Document detector rules

evaline-ju:
For: #193


gkumbhat (approved):
LGTM.

I have proposed a slight modification to the diagram. I'll push that update since Evaline is currently on vacation.

Issue #250, opened 2024-11-07
Integrate `tower::Service` into client infrastructure


pscoro:
## Description
In order to provide middleware (in particular for telemetry), we want to make use of `tower::Service` in our clients. We intend to wrap http and grpc client implementations each in tower services and enable middleware such as `tower_http::TraceLayer` to be used.

## Acceptance Criteria
- All http and grpc clients are layered tower services
- the client services properly support http, grpc, streaming and middleware while using tower
- client invocation is updated to execute through `service.call()`
- helper functions are added to http/grpc clients to simplify tower service calls

PR #259, opened 2024-12-05
:sparkles: Introduce tower service into clients

evaline-ju:
This PR adapts only the `tower::Service` integration from https://github.com/pscoro/fms-guardrails-orchestrator/tree/paul/tower-intro to:
- Add the `tower::Service` abstraction in client implementations so that tracing middleware can be added as a layer using one of many utilities built on top of the Service abstraction. This follows up to the #247 effort that migrated to `hyper` to enable more seamless `tower` usage.
- Replace the manual `tokio::timeout` `request_timeout` handling in favor of the `TimeoutLayer` from `tower`.
- Dependencies are updated to mainly use the latest `tower`. With the previous separate `tower-timeout` usage, etc. there were errors with tokio runtime that would appear on client health invocation.

For: #250

### DIFF
--- a/src/clients.rs
+++ b/src/clients.rs
@@ -312,6 +312,7 @@ pub async fn create_grpc_client<C>(
         .await
         .unwrap_or_else(|error| panic!("error creating grpc client: {error}"));
 
+    // Adds tower::Service wrapper to allow for enable middleware layers to be added
     let channel = ServiceBuilder::new().service(channel);
     new(channel)
 }


src/clients.rs
```diff
@@ -306,6 +311,8 @@ pub async fn create_grpc_client<C>(
         .channel()
         .await
         .unwrap_or_else(|error| panic!("error creating grpc client: {error}"));
+
+    let channel = ServiceBuilder::new().service(channel);
```
gkumbhat, line 315: why was this required?

gkumbhat (approved):
Looks good to me!

PR #260, opened 2024-12-10
:sparkles: Client metric trace layers

evaline-ju:
This PR adapts/builds on the trace layers in https://github.com/pscoro/fms-guardrails-orchestrator/tree/paul/trace-layer to introduce the client metrics collected per every http or grpc client request in a middleware layer on the service. This PR is dependent on/built off #259 , which introduces the `tower::Service` usage to enable middleware layers to be added.

From Paul's description:
- The [current] metrics [success/failure count, request durations] are fairly generic. Endpoint-specific metrics (if needed) can always be collected with the `info!` macros inside a given endpoint function. [This is enabled by the inclusion of the `tracing_opentelemetry` crate]
- The `TraceLayer` makes the underlying type of the client services much more complicated since `tower::Service` contains type generics, and associated types that get nested as layers are stacked

The `tracing_opentelemetry` crate expects histogram metrics to be prefixed with `histogram.` and counter metrics to be prefixed with `counter.` or `monotonic_counter.` The durations calculated with `as_millis()` have to be cast to `f/u64` for histograms, or they will not be logged remaining as `u128`.

If these updates are run with a collector, metrics can be found in prometheus [screenshot of metrics explore attached]
Screenshot 2024-12-10 at 10 46 10 AM

### DIFF
--- a/src/clients.rs
+++ b/src/clients.rs
@@ -498,11 +498,11 @@ impl OnFailure<GrpcFailureClass> for ClientOnFailure {
 
         let (status_code, error) = match failure_classification {
             GrpcFailureClass::Code(code) => {
-                error!(?trace_id, code, latency_ms, "failure handling request",);
+                error!(%trace_id, code, latency_ms, "failure handling request",);
                 (Some(grpc_to_http_code(tonic::Code::from(code.get()))), None)
             }
             GrpcFailureClass::Error(error) => {
-                error!(?trace_id, latency_ms, "failure handling request: {}", error,);
+                error!(%trace_id, latency_ms, "failure handling request: {}", error,);
                 (None, Some(error))
             }
         };


declark1 (commented):
IMO, we should have more concise trace messages, modeled after the default implementations. I would also recommend that we assign all details/metrics to fields rather than embedding them in the trace message; this is more user-friendly from the telemetry dashboard user perspective, e.g. I could easily build a chart showing average latency for a given method.

A few examples (current vs. recommended):

`OnRequest`:
```rust
info!(
    "outgoing HTTP client request to {} {} with trace {:?}",
    request.method(),
    request.uri().path(),
    trace::current_trace_id(),
);
```
```rust
info!(
    trace_id = %trace::current_trace_id(),
    method = %request.method(),
    path = %request.uri().path(),
    "started processing request",
);
```

`OnResponse`:
```rust
info!(
    "HTTP client response {} for request with trace {:?} received in {} ms",
    &response.status(),
    trace::current_trace_id(),
    latency.as_millis()
);
```
```rust
info!(
    trace_id = %trace::current_trace_id(),
    status = %response.status(),
    duration_ms = %latency.as_millis(),
    "finished processing request",
);
```

`OnEos`:
```rust
info!(
    "HTTP client stream response for request with trace {:?} closed after {} ms with trailers: {:?}",
    trace::current_trace_id(),
    stream_duration.as_millis(),
    trailers
);
```
```rust
info!(
    trace_id = %trace::current_trace_id(),
    duration_ms = stream_duration.as_millis(),
    "end of stream",
);
// NOTE: purposely excluded `trailers` as we probably don't need it here
```

src/clients/http.rs
```diff
@@ -271,23 +291,191 @@ impl HttpClient {
...
+                monotonic_counter.client_success_response_count = 1,
+                response_status = response.status().as_u16(),
+                request_duration = latency.as_millis()
+            );
+        } else {
+            error!(
+                "unexpected HTTP client response status code: {}",
+                response.status().as_u16()
+            );
+        }
```
declark1, line 414: I haven't tested the code, but this looks to be sending traces with duplicate data. 

The first trace, sent for every response, includes `response_status` and `request_duration` fields. The second trace, also sent for every response, also has a `response_status` field. I think the second trace can be dropped with `histogram.client_request_duration` moved up to the first trace, although I'm not familiar with the counters.

The if else blocks are also sending traces with `response_status` and `request_duration` fields rather than just incrementing their unique counters.

src/clients/http.rs
```diff
@@ -271,23 +291,191 @@ impl HttpClient {
...
+            monotonic_counter.client_request_failure_count = 1,
+            latency_ms,
+            ?status_code,
+            ?error
+        );
+        info!(
+            monotonic_counter.client_5xx_response_count = 1,
+            latency_ms,
+            ?status_code,
+            ?error
```
declark1, line 458: Duplicated fields, similar to comment above. Also, we should use the fields `Display` implementations (`%`) rather than `Debug` (`?`) in most cases.

src/clients/http.rs
```diff
@@ -271,23 +291,191 @@ impl HttpClient {
...
+        info!(
+            trace_id = %trace::current_trace_id(),
+            duration_ms = stream_duration.as_millis(),
+            "end of stream",
+        );
+        info!(
+            monotonic_counter.client_stream_response_count = 1,
+            stream_duration = stream_duration.as_millis()
+        );
+        info!(histogram.client_stream_response_duration = stream_duration.as_millis() as u64);
```
declark1, line 478: Duplicated fields, similar to comments above. I believe the counters can be moved up to the initial trace.

src/clients/http.rs
declark1: I used the wrong terminology. Each `info!()` or other `event!()` variant produces a new "event", not a new "trace", within the context of a "span".

declark1 (approved):
LGTM

PR #269, opened 2024-12-27
:recycle: Combine tracing info logs

evaline-ju:
In #260 , client tracing info logs were combined, but this was not refactored for the previously added logs for (overall) orchestrator requests. This PR refactors those logs


gkumbhat (approved):
Looks good to me!

PR #273, opened 2025-01-07
:fire: Remove legacy tokenization client

evaline-ju:
This client would've been for HTTP chunking and/or tokenization, which is currently not a priority, and this client is currently unused.


declark1 (approved):
LGTM

PR #276, opened 2025-01-10
Add chat content detector (input detection only)

gkumbhat:
### Changes

- Add `detectors` block in ChatCompletionRequest as per API documented [here](https://foundation-model-stack.github.io/fms-guardrails-orchestrator/#/Task%20-%20Chat%20Completions%2C%20with%20detection/api_v2_chat_completions_detection_handler)
- Add `detections` and `warnings` in `ChatCompletion` response object
- (Design) Add `detector_processing` folder under `src/orchestrator` that is intended to contain modules containing specific to particular detector containing functions required to do necessary processing for orchestration.
- Add an internal chat message representation called `ChatMessagesInternal`. This is to store chat messages in common format across all detector and chat completion processing. This will be used to processing chat messages both at request and response.
- (Design) Add content detector related processing containing function to filter out messages given chat input. This filtering function implements a rule that validates messages it can process and returns last message that the content detector can process. (Note that content detector can only process 1 message, since its designed to work on 1 text content)
- Add input detection (using detector of content type) in chat completion-detection handler that runs list of detectors provided as "input detectors" and returns detections to user, in case of any detections otherwise calls out to chat completion.

### Request
(with detection)

```
    curl -X 'POST' \
      'http://localhost:8038/api/v2/chat/completions-detection' \
      -H 'accept: application/json' \
      -H 'Content-Type: application/json' \
      -d '{
      "messages": [
        {
          "content": "this is a nice sentence",
          "role": "user",
          "name": "string"
        },
        {
          "content": "this is a stupid sentence",
          "role": "user",
          "name": "string"
        }
      ],
      "model": "llama-3",
      "n": 1,
      "temperature": 1,
      "top_p": 1,
      "user": "user-1234",
      "detectors": {
        "input": {
          "en_syntax_slate.38m.hap": {}
        }
      }
    }
```

### Response

```
    {
      "choices" : [],
      "created" : 1736551467,
      "detections" : {
        "input" : [
            {
              "message_index" : 2,
              "results" : [
                  {
                    "detection" : "has_HAP",
                    "detection_type" : "hap",
                    "end" : 25,
                    "score" : 0.984023928642273,
                    "start" : 0,
                    "text" : "this is a stupid sentence"
                  }
              ]
            }
        ]
      },
      "id" : "dummy-id",
      "model" : "llama-3",
      "object" : "",
      "usage" : {
        "completion_tokens" : 0,
        "prompt_tokens" : 0,
        "total_tokens" : 0
      }
    }
```


src/clients/openai.rs
```diff
@@ -665,3 +691,28 @@ pub struct OpenAiError {
...
+    pub message_index: u16,
+    pub results: Option<Vec<OrchestratorDetectionResult>>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct OutputDetectionResult {
+    choice_index: u16,
+    results: Option<Vec<OrchestratorDetectionResult>>,
+}
+
```
declark1, line 713: As this grows in complexity, we should carefully consider how we are naming objects to avoid ambiguity and confusion. We should also be consistent in naming conventions and types used for certain scenarios. 

For example, it is often better to _not_ wrap collections such as `Vec` in `Option` (improves ergonomics and `is_empty()` can be used instead of `is_some()` / `is_none()`), with `#[serde(default)]` for serializable types, and `usize` is idiomatic to use for "index" type values even if capacity is larger than needed. I also notice that we are using "Response" and "Result" interchangeably in the naming of objects and are overusing "result" in the naming of fields and objects.

(will add these notes to refactor document)

- In this case, I would recommend that we rename `DetectionResult` to something more specific as there is another type with this name. Since this is a container holding input and output detections and assigned to a field named `detections`, `ChatDetections` would probably be a good name.
- I recommend dropping `Option` from the fields and replacing `#[serde(skip_serializing_if = "Option::is_none")]` with`#[serde(default, skip_serializing_if = "Vec::is_empty")]`
- Do we really need separate `InputDetectionResult` and `OutputDetectionResult` types? It feels like we could have a single type with an `index` field used for both
- `OrchestratorDetectionResult` also needs a better name. If feels weird to have a "DetectionResult" type with a `results` field containing a vec of `OrchestratorDetectionResult`. Similar to above, would should drop the `Option` from the `results` field

src/orchestrator/chat_completions_detection.rs
```diff
@@ -1,27 +1,487 @@
...
+    pub role: String,
+    /// The contents of the message.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub content: Option<Content>,
+    /// The refusal message by the assistant. (assistant message only)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub refusal: Option<String>,
+}
+
+pub type ChatMessagesInternal = Vec<ChatMessageInternal>;
```
declark1, line 67: nit: I recommend that we just use `Vec<ChatMessageInternal>` instead of a type alias so that it is clear it is a collection of messages

src/orchestrator/chat_completions_detection.rs
```diff
@@ -1,27 +1,487 @@
...
+                    message_index: index,
+                    role: m.role.clone(),
+                    content: m.content.clone(),
+                    refusal: m.refusal.clone(),
+                }
+            })
+        });
+        messages
+    }
+}
```
declark1, line 89: This is inefficient; allocates every `for_each` iteration as `Vec::new()` was used instead of `Vec::with_capacity(n)` and it creates unnecessary copies of the message contents. Use `into_iter()` and `map()`:

```suggestion
impl From<ChatCompletionsRequest> for ChatMessagesInternal {
    fn from(value: ChatCompletionsRequest) -> Self {
        value
            .messages
            .into_iter()
            .enumerate()
            .map(|(index, message)| ChatMessageInternal {
                message_index: index,
                role: message.role,
                content: message.content,
                refusal: message.refusal,
            })
            .collect()
    }
}
```

src/orchestrator/chat_completions_detection.rs
```diff
@@ -1,27 +1,487 @@
...
+                                Ok(value) => {
+                                    if !value.is_empty() {
+                                        let input_detection_result = InputDetectionResult {
+                                            message_index: idx as u16,
+                                            results: Some(
+                                                value
+                                                .iter()
+                                                .map(|result| {OrchestratorDetectionResult::ContentAnalysisResponse(result.clone())})
+                                                .collect::<Vec<_>>())
+                                        };
```
declark1, line 271: Performs unnecessary copies of owned values. Use `into_iter()` and drop the clone. Can be written more concisely:
`results: Some(value.into_iter().map(OrchestratorDetectionResult::ContentAnalysisResponse).collect())`

(side note mentioned above: we should drop the `Option`)

src/clients/openai.rs
```diff
@@ -157,13 +161,13 @@ impl HttpClientExt for OpenAiClient {
 
 #[derive(Debug)]
 pub enum ChatCompletionsResponse {
-    Unary(ChatCompletion),
+    Unary(Box<ChatCompletion>),
```
declark1, line 164: Why is this boxed?

src/orchestrator/chat_completions_detection.rs
```diff
@@ -1,27 +1,487 @@
...
+                }
+            }
+        }))
+        .await
+        .into_iter()
+        .collect::<Result<Vec<_>, Error>>();
+
+        match results {
+            Ok(chunk_value) => {
+                let chunk_values = chunk_value.into_iter().collect::<Vec<_>>();
```
declark1, line 389: `chunk_values` creates a duplicate collection of the same type

src/orchestrator/chat_completions_detection.rs
```diff
@@ -1,27 +1,487 @@
...
+            if let Some(mut input_detections) = input_detections {
+                // Sort input detections by message_index
+                input_detections.sort_by_key(|value| value.message_index);
+
+                let input_detections = input_detections
+                    .into_iter()
+                    .map(|detection| {
+                        let last_idx = detection.results.clone().unwrap().len();
+                        // sort detection by starting span, if span is not present then move to the end of the message
+                        detection.results.clone().unwrap().sort_by_key(|r| match r {
```
declark1, line 141: This is creating 2 extra clones of the detection results.

src/orchestrator/chat_completions_detection.rs
```diff
@@ -1,27 +1,487 @@
...
+    orchestrator::{
+        detector_processing::content,
+        unary::{chunk, detect_content},
+        Chunk,
+    },
+};
+use serde::{Deserialize, Serialize};
+use uuid::Uuid;
+
+pub type ChunkResult<T> = Pin<Box<dyn Future<Output = T> + Send>>;
```
declark1, line 49: It looks like this is unused, so it should be dropped. The futures crate has a type alias for owned dynamically typed futures `futures::future::BoxFuture`.

src/orchestrator/detector_processing/content.rs
```diff
@@ -0,0 +1,157 @@
...
+                "Message at last index is not from user or assistant".into(),
+            ))
+        }
+    }
+
+    let content = match message.content.clone().unwrap() {
+        Content::Text(text) => Content::Text(text),
+        _ => return Err(ValidationError::Invalid("Incorrect type requested".into())),
+    };
+    Ok(ChatMessagesInternal::from(vec![ChatMessageInternal {
```
declark1, line 55: Why do we build a vec if only 1 message is supported?

src/clients/openai.rs
gkumbhat: this came out as a suggestion from clippy.

```
consider boxing the large fields to reduce the total size of the enum
```

src/clients/openai.rs
```diff
@@ -275,6 +279,22 @@ pub struct ChatCompletionsRequest {
     pub skip_special_tokens: Option<bool>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub spaces_between_special_tokens: Option<bool>,
+
+    // Detectors
+    // Note: We need to make it optional, since this structure also gets used to
+    // form request for chat completions. And downstream server, like vLLM would
+    // reject extra parameters.
```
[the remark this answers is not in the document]
gkumbhat: I'll update the comment

src/orchestrator/chat_completions_detection.rs
gkumbhat: good catch. Will remove it. And thanks for the BoxFuture reference, seems helpful for future.

src/orchestrator/detector_processing/content.rs
```diff
@@ -0,0 +1,157 @@
...
+) -> Result<ChatMessagesInternal, ValidationError> {
+    // Implement content processing logic here
+    // Rules:
+    // Rule 1: Select last message from the list of messages
+    let message = messages.last().unwrap();
+
+    // Rule 2: Check if the message has content or not
+    if message.content.is_none() {
+        return Err(ValidationError::Invalid(
+            "Message at last index does not have content".into(),
```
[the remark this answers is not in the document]
gkumbhat: Used the term filtering for more general sense, since for other detector types, like chat one's, we would be filtering out messages of role's not compatible (atleast with existing version of detector).

So to be able to refer to function with same name across different detector type (in their own module), I used the name "filtering".

src/orchestrator/detector_processing/content.rs
```diff
@@ -0,0 +1,157 @@
...
+    Ok(ChatMessagesInternal::from(vec![ChatMessageInternal {
+        // index of last message
+        message_index: messages.len() - 1,
+        role: message.role.clone(),
+        content: Some(content),
+        refusal: message.refusal.clone(),
+    }]))
+}
+
+pub async fn get_content_analysis_request(
```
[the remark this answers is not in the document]
gkumbhat: this was used in early iterations, not used anymore. I'll remove it.

src/clients/openai.rs
gkumbhat: > In this case, I would recommend that we rename DetectionResult to something more specific as there is another type with this name. Since this is a container holding input and output detections and assigned to a field named detections on the chat response type, ChatDetections would probably be a good name.

Sure. Will make the change.


> I recommend dropping Option from the fields and replacing #[serde(skip_serializing_if = "Option::is_none")] with #[serde(default, skip_serializing_if = "Vec::is_empty")]

Sounds good. I'll make the change

> Do we really need separate InputDetectionResult and OutputDetectionResult types?

Yes, this was added to particularly specify that in output, we are talking about `choice_index` since, since, choice is something only present in output but messages are present technically in both input and output.


> OrchestratorDetectionResult also needs a better name. It feels weird to have a "DetectionResult" type with a results field containing a vec of OrchestratorDetectionResult. 

How about: `GuardrailDetection` ?


> Similar to above, we should drop the Option from the results field.

src/orchestrator/chat_completions_detection.rs
```diff
@@ -1,27 +1,487 @@
...
+// Function to chunk ChatMessagesInternal based on the chunker id and return chunks in ChatMessagesInternal form
+// Output maps each detector_id with corresponding chunk
+async fn detector_chunk_task(
+    ctx: &Arc<Context>,
+    detector_chat_messages: HashMap<String, ChatMessagesInternal>,
+) -> Result<HashMap<String, Vec<(usize, Vec<Chunk>)>>, Error> {
+    let mut chunks = HashMap::<String, Vec<(usize, Vec<Chunk>)>>::new();
+
+    // TODO: Improve error handling for the code below
+    for (detector_id, chat_messages) in detector_chat_messages.iter() {
```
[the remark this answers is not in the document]
gkumbhat: yes, that is correct. We can optimize duplicate chunking later on, but since that can make it more complex, I tried to not go in that direction yet in this PR.

src/clients/openai.rs
declark1: Hmm, thinking about this further, I highly recommend that we resolve the detection result type inconsistency to make this cleaner here and elsewhere.

Currently, all detectors return `Vec<DetectionResult>` except for text contents which returns `Vec<Vec<ContentAnalysisResponse>>`, requiring us to add in this additional enum in an already multiple-layer result type.
- Can't recall why its a nested vec for text contents

For this scenario, you would typically have a `DetectionKind` enum or similar with a variant for each type like this:

```rust
pub enum DetectionKind {
    TextChat(TextChatDetection),
    TextContents(TextContentsDetection),
    TextContextDoc(TextContextDocDetection),
    TextGeneration(TextGenerationDetection),
}
```
In our case, we would end up with the following, which is pretty odd:
```rust
pub enum DetectionKind {
    ContentAnalysis(ContentAnalysisResponse),
    DetectionResult(DetectionResult),
}
```

We should also avoid using "result" in the naming of types (for most cases) as `Result` is an enum holding success/error variants; there is usually a better alternative.

I would suggest the following types except ideally replacing `DetectionKind` with `DetectionResult`. If its alright, I can look into this to confirm if it will work and open a separate PR to make the required changes to update text contents detector.

```rust
#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct ChatDetections {
    #[serde(default, skip_serializing_if = "Vec::is_empty")]
    pub input: Vec<InputChatDetection>,
    #[serde(default, skip_serializing_if = "Vec::is_empty")]
    pub output: Vec<OutputChatDetection>,
}

#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct InputChatDetection {
    pub message_index: usize,
    pub results: Vec<DetectionKind>, // ideally Vec<DetectionResult>,
}

#[derive(Debug, Clone, Serialize, Deserialize)]
pub struct OutputChatDetection {
    pub choice_index: usize,
    pub results: Vec<DetectionKind>, // ideally Vec<DetectionResult>,
}

#[derive(Debug, Clone, Serialize, Deserialize)]
#[serde(untagged)]
pub enum DetectionKind {} // ideally use `DetectionResult` instead
```

src/orchestrator/detector_processing/content.rs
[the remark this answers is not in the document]
declark1: What different filtering is required for other detector types? Will other detector types actually be applicable to chat? Just trying to understand the need for modules per detector type.

src/orchestrator/detector_processing/content.rs
```diff
@@ -0,0 +1,129 @@
...
+};
+
+/// Function to get content analysis request from chat message by applying rules
+pub fn filter_chat_message(
+    messages: Vec<ChatMessageInternal>,
+) -> Result<Vec<ChatMessageInternal>, ValidationError> {
+    // Implement content processing logic here
+    // Rules:
+    // Rule 1: Select last message from the list of messages
+    let message = messages.last().unwrap();
```
declark1, line 29: We should probably check `messages.is_empty()` and return `ValidationError`, to ensure that `messages.last().unwrap()` doesn't panic.

src/clients/openai.rs
gkumbhat: Agree, this looks a bit messy. 

I like each kind having its own type. That will be cleaner..

There are a couple of downside of going this route as well:
1.  Having different types would mean there can be potential of inconsistency between them to refer to same logical element. I am more referring to things that are more development error. For example, one can call `explanation` vs `explanations`.. However, these are something we can be careful about in reviews, so avoidable.
2. For complex endpoints, such as `chat` we would end up doing lot of `match` operator, to put output in appropriate data types.  I feel, this is avoidable as well with some handy utility functions associated with each of these "kinds". 

Since this change would require changes in other places in code, I would like to move this into its own effort / issue and get it in as a separate PR.  Also, while adding this, it would be good to factor in points mentioned above and see how we can programmatically make it easier to resolve those.

src/orchestrator/chat_completions_detection.rs
```diff
@@ -1,27 +1,487 @@
...
+                    model: request.model,
+                    choices: vec![],
+                    created: SystemTime::now()
+                        .duration_since(UNIX_EPOCH)
+                        .unwrap()
+                        .as_secs() as i64,
+                    detections: Some(DetectionResult {
+                        input: Some(input_detections),
+                        output: None,
+                    }),
```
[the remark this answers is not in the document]
gkumbhat: Ah yes, I missed it. I'll add the warning

src/clients/openai.rs
declark1: Got it. This is because the size of an enum is equal to the size of it's largest variant. Since the `Streaming` variant holds `mpsc::Receiver<T>` which is pointer-sized (16 bytes), it is much smaller than the `Unary` variant which is the size of the `ChatCompletion` struct (264 bytes). Boxing it makes it a pointer to `ChatCompletion`, so this reduces the size of the enum to 16 bytes.

src/orchestrator/chat_completions_detection.rs
```diff
@@ -125,12 +125,15 @@ impl Orchestrator {
...
                         // sort detection by starting span, if span is not present then move to the end of the message
-                        detection.results.clone().sort_by_key(|r| match r {
-                            GuardrailDetection::ContentAnalysisResponse(value) => value.start,
-                            _ => last_idx,
+                        detection.results.sort_by(|a, b| match (a, b) {
+                            (
+                                GuardrailDetection::ContentAnalysisResponse(content_a),
+                                GuardrailDetection::ContentAnalysisResponse(content_b),
+                            ) => content_a.start.cmp(&content_b.start),
+                            _ => last_idx.cmp(&last_idx),
```
declark1, line 136: There is still an unneccessary clone on line 129 to get `last_idx`. The previous `sort_by_key` approach seemed simpler; is there any difference using `sort_by` here?

src/orchestrator/chat_completions_detection.rs
```diff
@@ -1,27 +1,478 @@
...
+                let detector_config = ctx
+                    .config
+                    .detectors
+                    .get(&detector_id)
+                    .unwrap_or_else(|| panic!("detector config not found for {}", detector_id));
+                let detector_type = &detector_config.r#type;
+                // Filter messages based on detector type
+                let value = match detector_type {
+                    DetectorType::TextContents => {
+                        match content::filter_chat_message(messages.clone()) {
```
declark1, line 316: It seems wasteful to clone `messages` here, just to take a reference to the last message and clone it again. I would update `filter_chat_message` to take a slice instead and drop the clone here.

A more Rusty approach using iterators and iterator methods would be better IMO, but I can't really make a recommendation without understanding where we are going with this.

src/orchestrator/chat_completions_detection.rs
gkumbhat: Missed line 129 (will fix that)..

I just tried this `sort_by` approach. Looked like equally efficient 🤔

src/orchestrator/chat_completions_detection.rs
```diff
@@ -1,27 +1,478 @@
...
+
+    // TODO: Improve error handling for the code below
+    for (detector_id, chat_messages) in detector_chat_messages.iter() {
+        let chunk_tasks = chat_messages
+            .iter()
+            .map(|message| {
+                let text = match message.content.as_ref().unwrap() {
+                    Content::Text(value) => value,
+                    _ => panic!("Only text content accepted"),
+                };
```
declark1, line 345: `detector_chunk_task` receives owned values, so you can change both `iter()` to `into_iter()`, drop the `.as_ref()` on line 350 and subsequent `text.clone()` on line 357.

src/orchestrator/chat_completions_detection.rs
```diff
@@ -1,27 +1,478 @@
...
+        .collect::<Result<HashMap<String, Vec<ChatMessageInternal>>, Error>>()
+}
+
+// Function to chunk Vec<ChatMessageInternal> based on the chunker id and return chunks in Vec<ChatMessageInternal> form
+// Output maps each detector_id with corresponding chunk
+async fn detector_chunk_task(
+    ctx: &Arc<Context>,
+    detector_chat_messages: HashMap<String, Vec<ChatMessageInternal>>,
+) -> Result<HashMap<String, Vec<(usize, Vec<Chunk>)>>, Error> {
+    let mut chunks = HashMap::<String, Vec<(usize, Vec<Chunk>)>>::new();
```
declark1, line 335: nit: rust has type inference
```suggestion
    let mut chunks = HashMap::new();
```

src/orchestrator/chat_completions_detection.rs
```diff
@@ -1,27 +1,478 @@
...
+                            Ok(filtered_messages) => Ok(filtered_messages),
+                            Err(e) => Err(Error::Validation(e.to_string())),
+                        }
+                    }
+                    _ => unimplemented!(),
+                }?;
+                Ok((detector_id, value))
+            },
+        )
+        .collect::<Result<HashMap<String, Vec<ChatMessageInternal>>, Error>>()
```
declark1, line 326: nit: type is inferred from return type
```suggestion
        .collect()
```

src/orchestrator/chat_completions_detection.rs
declark1: I haven't got into using it much yet since stabilized, but seems like a good `let-else` use case here.

```rust
let Some(Content::Text(text)) = message.content else {
    panic!("Only text content accepted")
};
```

src/orchestrator/detector_processing/content.rs
```diff
@@ -0,0 +1,134 @@
...
+        }
+    };
+    Ok(vec![ChatMessageInternal {
+        // index of last message
+        message_index: messages.len() - 1,
+        role: message.role.clone(),
+        content: Some(content),
+        refusal: message.refusal.clone(),
+    }])
+}
```
declark1, line 68: This is more of a nit, but I would consolidate validating "text content" as a single rule, rather than one for content and one for text. Also would leverage the `match!` macro for pattern match validations (rust style preference). I also cloned the message once instead of cloning it's individual fields and slightly reworded the error messages and comments.

I noticed that the function is called `filter_chat_message` rather than `filter_chat_messages`, even though it returns a vec. Are you planning to eventually support more than one message here?

```suggestion
pub fn filter_chat_message(
    messages: &[ChatMessageInternal],
) -> Result<Vec<ChatMessageInternal>, ValidationError> {
    // Get last message
    if messages.is_empty() {
        return Err(ValidationError::Invalid("No messages provided".into()));
    }
    let message_index = messages.len() - 1;
    let message = messages.last().unwrap().clone();

    // Validate message:
    // 1. Has text content
    if !matches!(message.content, Some(Content::Text(_))) {
        return Err(ValidationError::Invalid(
            "Last message content must be text".into(),
        ));
    }
    // 2. Role is user | assistant | system
    if !matches!(message.role.as_str(), "user" | "assistant" | "system") {
        return Err(ValidationError::Invalid(
            "Last message role must be user, assistant, or system".into(),
        ));
    }

    Ok(vec![ChatMessageInternal {
        message_index,
        role: message.role,
        content: message.content,
        refusal: message.refusal,
    }])
}
```

Note: if used, would require updating one of the error messages in the unit tests.

src/orchestrator/chat_completions_detection.rs
```diff
@@ -1,27 +1,476 @@
...
+                let detector_type = &detector_config.r#type;
+                // Filter messages based on detector type
+                let value = match detector_type {
+                    DetectorType::TextContents => match content::filter_chat_message(&messages) {
+                        Ok(filtered_messages) => Ok(filtered_messages),
+                        Err(e) => Err(Error::Validation(e.to_string())),
+                    },
+                    _ => unimplemented!(),
+                }?;
+                Ok((detector_id, value))
```
declark1, line 323: `From<ValidationError> for Error` is implemented, so the nested match can be dropped here as it will be implicitly converted. nit: Renamed `value` to `messages`.

```suggestion
                let messages = match detector_type {
                    DetectorType::TextContents => content::filter_chat_message(&messages),
                    _ => unimplemented!(),
                }?;
                Ok((detector_id, messages))
```

src/orchestrator/detector_processing/content.rs
gkumbhat: > I noticed that the function is called filter_chat_message rather than filter_chat_messages

The name for this function is like this because I had this in other place which was doing singular. The reason it returns vector is so that other detectors can have function named same but they would return more than 1 messages. Like the `chat` ones.

I'll update the name of the function

gkumbhat:
@declark1 pushed the fixes to test and renamed the function

src/orchestrator/chat_completions_detection.rs
```diff
@@ -1,27 +1,473 @@
...
-            .map_err(|error| Error::ChatGenerateRequestFailed {
-                id: model_id,
-                error,
+        let ctx = self.ctx.clone();
+        let headers = task.headers.clone();
+
+        let request = task.request.clone();
+        let task_handle = tokio::spawn(async move {
+            // Convert the request into a format that can be used for processing
+            let chat_messages = Vec::<ChatMessageInternal>::from(request.clone());
```
declark1, line 110: Currently, `ChatCompletionsRequest` gets cloned twice for each request (lines 107 and 110). Since this is a large struct (1168 bytes), we should try to avoid this. Ideally, we should convert `Message` into `ChatMessageInternal` without cloning and consuming `ChatCompletionsRequest` as the remaining fields are discarded. I can look at this further and provide some options unless you want to give it a shot first.

The `headers` clones can be avoided as well.

declark1:
> Add an internal chat message representation called ChatMessagesInternal. This is to store chat messages in common format across all detector and chat completion processing. This will be used to processing chat messages both at request and response.

Looking at this further (while working on refactor plan), I'm not seeing the need for `ChatMessageInternal`; IMO, it would be better to have `filter_chat_messages()` take a reference to `ChatCompletionsRequest`, apply validation/filtering logic, and return `Vec<(usize, String)`, i.e. (message_index, text) tuples. It would also solve one of the issues in my last comment above. I think this would be more efficient and also enable us to call a common/reusable chunk task (once implemented) instead of having a separate `detector_chunk_task()`here (would still need it for now).

src/orchestrator/chat_completions_detection.rs
gkumbhat: Pushed some changes to avoid cloning of entire request and headers.

src/orchestrator/detector_processing/content.rs
```diff
@@ -0,0 +1,119 @@
...
+ See the License for the specific language governing permissions and
+ limitations under the License.
+
+*/
+use crate::{
+    clients::openai::Content, models::ValidationError,
+    orchestrator::chat_completions_detection::ChatMessageInternal,
+};
+
+/// Function to get content analysis request from chat message by applying rules
```
declark1, line 22: Suggest rewording this doc string as it will be applied to representations of request and response messages.

declark1 (approved):

PR #278, opened 2025-01-15
Add detector_id in response for orchestrator apis

swith004:
**Story:**  ai-foundation/watson-fm-stack-tracker#1335

## What does your PR do?
This PR changes the models for `DetectionResponse`, `TokenClassificationResponse`, and `ContentAnalysisResponse` to include a detector_id along with the results, so that users can identify which detector provided the results (for instance, if there are multiple hap models, such as the cpu optimized version).

### How this was tested 
This was tested locally, by modifying the config yaml to account for local certs, port-forwarding of os-detectors deployed on openshift, and locally run detectors (for hap). The following are the curl commands used to verify the changes to the orchestrator: 

### api/v2/text/generation-detection
```
curl --request POST   --url 'http://localhost:8033/api/v2/text/generation-detection'  --header 'Content-Type: application/json'  --data '{"model_id": "bloom-560m",    "prompt": "All people from Kish are cruel and evil.","detectors": {       "os-detector-answer-relevance-v1": {}},"text_gen_parameters": {"preserve_input_text": true,"max_new_tokens": 99,"min_new_tokens": 1,"truncate_input_tokens": 500,"decoding_method": "SAMPLING","top_k": 2,"top_p": 0.9,"typical_p": 0.5,"temperature": 0.8,"seed": 42,"repetition_penalty": 2,"max_time": 0,"stop_sequences": ["42"]}}'

{"generated_text":"All people from Kish are cruel and evil. They have no respect for human life, they hate the innocent.\nI am a man who has been in this world to help others but I do not want anyone else suffering because of my actions or words that hurt them more than me\nThe only thing you can say is “you can’t be so bad” as if it were true! You shouldn’t even try!\nYou don’t need any kind support when you’re alone with your problems…just let yourself go through whatever it’s like without thinking about what’s","detections":[{"detection_type":"answer_relevance","detection":"relevant","detector_id":"os-detector-answer-relevance-v1","score":0.16986796259880066}],"input_token_count":9}
```

### api/v2/text/detection/generated
```
curl -k --url http://localhost:8033/api/v2/text/detection/generated  --header 'Content-Type: application/json' --data '{ "prompt": "What is the name of the solar powered car race held every two years?", "generated_text": "The name of the solar powered car race held every two years is The World Solar Challenge.", "detectors": {  "os-detector-answer-relevance-v1": {} }}'

{"detections":[{"detection_type":"answer_relevance","detection":"relevant","detector_id":"os-detector-answer-relevance-v1","score":0.995817482471466}]}
```
### api/v2/text/detection/context
```
curl --request 'POST'   --url 'http://localhost:8033/api/v2/text/detection/context'   --header 'Content-Type: application/json'   --data '{      "detectors": {        "os-detector-context-relevance-v1": {}      },      "content": "What is the name of the solar powered car race held every two years?",      "context_type": "url",      "context": [        "https://en.wikipedia.org/wiki/IBM",        "https://research.ibm.com/"      ]}'

{"detections":[{"detection_type":"context_relevance","detection":"relevant","detector_id":"os-detector-context-relevance-v1","score":0.7458338141441345,"evidence":[{"name":"context","score":0.6409949064254761,"evidence":[{"name":"context_chunk","value":"https://en.wikipedia.org/wiki/IBM","score":0.6409949064254761}]},{"name":"context","score":0.7458338141441345,"evidence":[{"name":"context_chunk","value":"https://research.ibm.com/","score":0.7458338141441345}]}]}]}
```

### api/v2/text/detection/content
```
curl --request POST     --url 'http://localhost:8033/api/v2/text/detection/content'     --header 'accept: application/json'     --header 'Content-Type: application/json'     --data '{
      "detectors": {
        "en_syntax_slate.38m.hap": {}
      },
      "content": "All people from Kish are stupid and ignorant."
}'

{"detections":[{"start":0,"end":45,"text":"All people from Kish are stupid and ignorant.","detection":"has_HAP","detection_type":"hap","detector_id":"en_syntax_slate.38m.hap","score":0.9972074627876282}]}
```

### api/v1/task/classification-with-text-generation
```
curl --request POST   --url 'http://localhost:8033/api/v1/task/classification-with-text-generation'  --header 'Content-Type: application/json'  --data '{"inputs": "All people from Kish are cruel and evil.","model_id": "bloom-560m","guardrail_config": {"input": {"models": {"en_syntax_slate.38m.hap": {}}},"output": {"models": {"en_syntax_slate.38m.hap": {}}}},"text_gen_parameters": {"preserve_input_text": true,"max_new_tokens": 99,"min_new_tokens": 1,"truncate_input_tokens": 500,"decoding_method": "SAMPLING","top_k": 2,"top_p": 0.9,"typical_p": 0.5,"temperature": 0.8,"seed": 42,"repetition_penalty": 2,"max_time": 0,"stop_sequences": ["42"]}}'

{"token_classification_results":{"input":[{"start":0,"end":40,"word":"All people from Kish are cruel and evil.","entity":"has_HAP","entity_group":"hap","detector_id":"en_syntax_slate.38m.hap","score":0.9853121638298036}]},"input_token_count":9,"warnings":[{"id":"UNSUITABLE_INPUT","message":"Unsuitable input detected. Please check the detected entities on your input and try again with the unsuitable input removed."}]}
```

### api/v1/task/server-streaming-classification-with-text-generation
```
curl --request POST   --url 'http://localhost:8033/api/v1/task/server-streaming-classification-with-text-generation'  --header 'Content-Type: application/json'  --data '{"inputs": "All people from Kish are cruel and evil.","model_id": "bloom-560m","guardrail_config": {"input": {"models": {"en_syntax_slate.38m.hap": {}}},"output": {"models": {"en_syntax_slate.38m.hap": {}}}},"text_gen_parameters": {"preserve_input_text": true,"max_new_tokens": 99,"min_new_tokens": 1,"truncate_input_tokens": 500,"decoding_method": "SAMPLING","top_k": 2,"top_p": 0.9,"typical_p": 0.5,"temperature": 0.8,"seed": 42,"repetition_penalty": 2,"max_time": 0,"stop_sequences": ["42"]}}'

data: {"token_classification_results":{"input":[{"start":0,"end":40,"word":"All people from Kish are cruel and evil.","entity":"has_HAP","entity_group":"hap","detector_id":"en_syntax_slate.38m.hap","score":0.9853121638298036}]},"input_token_count":9,"warnings":[{"id":"UNSUITABLE_INPUT","message":"Unsuitable input detected. Please check the detected entities on your input and try again with the unsuitable input removed."}]}
```


protos/caikit_data_model_nlp.proto
```diff
@@ -89,8 +89,9 @@ message TokenClassificationResult {
   string word = 3;
   string entity = 4;
   string entity_group = 5;
-  double score = 6;
-  int64 token_count = 7;
+  string detector_id = 6;
```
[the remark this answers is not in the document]
swith004: I know it wasn't required for this change, but wasn't sure if this tokenclassificationresult differing from the other in models.rs would have unintended consequences. In that case, should I just revert back for this PR?

src/orchestrator/unary.rs
```diff
@@ -844,6 +856,16 @@ pub async fn detect_for_chat(
...
                 .into_iter()
                 .filter(|detection| detection.score > threshold)
+                .map(|detection| {
+                    //add detector_id
+                    DetectionResult {
+                        detection_type: detection.detection_type,
+                        detection: detection.detection,
+                        detector_id: Some(detector_id.clone()),
+                        score: detection.score,
+                        evidence: detection.evidence,
```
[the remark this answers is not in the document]
declark1: Since you are iterating over owned values of `DetectionResult`, you can just make it mutable and update the `detector_id` field, e.g.

```rust
// ...
.map(|mut detection| {
    // Set detector_id
    detection.detector_id = Some(detector_id.clone());
    detection
})
// ...
```

src/clients/detector/text_contents.rs
```diff
@@ -132,22 +132,26 @@ pub struct ContentAnalysisResponse {
...
     /// Score of detection
     pub score: f64,
     /// Optional, any applicable evidence for detection
     #[serde(skip_serializing_if = "Option::is_none")]
     pub evidence: Option<Vec<EvidenceObj>>,
 }
 
-impl From<ContentAnalysisResponse> for crate::models::TokenClassificationResult {
-    fn from(value: ContentAnalysisResponse) -> Self {
+impl From<(String, ContentAnalysisResponse)> for crate::models::TokenClassificationResult {
```
declark1, line 144: Since `detector_id` is being added to both types, is this needed? It appears that adding the field to the existing `From` implementation would be sufficient rather than using a tuple to provide it.

protos/caikit_data_model_nlp.proto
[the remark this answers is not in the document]
swith004: discussed this in standup - reverted back to original version in latest push

src/clients/detector/text_contents.rs
swith004: Here we do, because in the detect function in unary.rs, the ContentAnalysisResponse that gets returned from the request don't have a detector_id at this point in time, and once they are filtered to only get the responses that had a score exceeding some threshold, we take those results and will need to populate the resulting TokenClassifictionResults with detector_id (passed in as a tuple).

declark1: Hmm, I see. This is actually why I feel the `detector_id` should be added when the `DetectionResult` is created by the client (it is an argument to the client methods anyway), rather than added after the fact.

However, as a generic `post_to_detector` function was added recently, it makes setting it here less straight forward.

declark1: For the approach mentioned above, we would have to do something like this for each detector client.

1. Use the "turbofish" operator (`::<T>`) to specify the concrete return type `T` as `post_to_detector` is generic over `T`.
2. Use `Result::map()` to set `detector_id` for each item:

```rust
// Example using TextChatDetectorClient
    #[instrument(skip_all, fields(model_id, ?headers))]
    pub async fn text_chat(
        &self,
        model_id: &str,
        request: ChatDetectionRequest,
        headers: HeaderMap,
    ) -> Result<Vec<DetectionResult>, Error> {
        let url = self.endpoint(CHAT_DETECTOR_ENDPOINT);
        info!("sending text chat detector request to {}", url);
        self.post_to_detector::<Vec<DetectionResult>>(model_id, url, headers, request)
            .await
            .map(|detections| {
                detections
                    .into_iter()
                    .map(|mut detection| {
                        detection.detector_id = Some(model_id.to_string());
                        detection
                    })
                    .collect()
            })
    }
```

swith004: Hmm, I can review the code base again and look for a better place to populate the detector_id if after filtering the contentanalsyisresponse isn't ideal...

declark1: I just wanted to share the example above so that you know how you could do this, but it feels like it should really be returned by the detectors IMO. I would just leave what you have for now.

declark1: I just wanted to share the example above so that you know how you could set detector_id when created by the client (before applying filters, then you also wouldn't need the tuple for the From conversion), but it feels like it should really be returned by the detectors IMO. I would just leave what you have for now.

src/orchestrator/unary.rs
```diff
@@ -695,9 +695,10 @@ pub async fn detect(
             response
                 .into_iter()
                 .filter_map(|resp| {
-                    let mut result: TokenClassificationResult = resp.into();
+                    let mut result: TokenClassificationResult = (detector_id.clone(), resp).into();
                     result.start += chunk.offset as u32;
                     result.end += chunk.offset as u32;
```
declark1, line 700: For consistency with how `detector_id` is set on `DetectionResult` and to avoid the weird tuple `From` impl, I would revert the `From` impl change and just update the field like below:

e.g.
```suggestion
                    let mut result: TokenClassificationResult = resp.into();
                    result.detector_id = Some(detector_id.clone());
                    result.start += chunk.offset as u32;
                    result.end += chunk.offset as u32;
```

src/orchestrator/unary.rs
```diff
@@ -696,6 +696,7 @@ pub async fn detect(
                 .into_iter()
                 .filter_map(|resp| {
                     let mut result: TokenClassificationResult = resp.into();
+                    result.detector_id = Some(detector_id.clone()); // attach detector_id to the result, as ContentAnalsyisResponse does not contain one at this point
```
declark1, line 699: nit: use consistent comments throughout if they are needed. `// add detector_id` is sufficient and should be placed on the line above in all instances.

swith004: thanks. addressed in latest push

declark1 (approved):
LGTM

Issue #274, opened 2025-01-09
Update dependencies and drop unused dependencies


declark1:
## Description
Dependencies need to be updated.

## Acceptance Criteria
- [ ] Dependencies have been updated to latest versions in `Cargo.toml` and `Cargo.lock`
- [ ] Unused dependencies have been dropped
- [ ] prost_build `build_server` option has been disabled as only clients are needed (in `build.rs`)
- [ ] Integration tests pass successfully

PR #281, opened 2025-01-16
:arrow_up: Upgrade Rust and project dependencies

evaline-ju:
- Upgrade Rust and protoc versions
- Linting update due to upgraded Rust and removal of duplicate test
- Upgrade most dependencies and remove unused ones that are not explicitly used (async-stream, eventsource, futures-core, futures-timer, mime, mio, uuid)
- OpenTelemetry-related dependencies will need more investigation as the latest versions include breaking changes

Part of: #274

### DIFF
--- a/src/orchestrator/streaming/aggregator.rs
+++ b/src/orchestrator/streaming/aggregator.rs
@@ -592,9 +592,9 @@ mod tests {
 
         // We do not have all detections for the first chunk
         assert!(
-            !tracker
+            tracker
                 .first()
-                .is_some_and(|first| first.detections.len() == n_detectors),
+                .is_none_or(|first| first.detections.len() != n_detectors),
             "detections length should not be 2 for first chunk"
         );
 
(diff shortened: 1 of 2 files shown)


declark1:
We should update Rust too in `rust-toolchain.toml`. The latest stable is 1.84.0, we are on 1.80.1.

src/orchestrator/streaming/aggregator.rs
```diff
@@ -592,9 +592,9 @@ mod tests {
 
         // We do not have all detections for the first chunk
         assert!(
-            !tracker
+            tracker
                 .first()
-                .is_some_and(|first| first.detections.len() == n_detectors),
+                .is_none_or(|first| first.detections.len() != n_detectors),
```
declark1, line 597: This must be new in 1.84, never seen `Option::is_none_or` used

declark1 (approved):
LGTM

PR #282, opened 2025-01-17
:arrow_up::recycle: Update open telemetry dependencies

evaline-ju:
Updating to the latest open telemetry-related crates, particularly centering around use of `opentelemetry 0.27.0`

- The main interface changes are to the pipeline concept, instead switching to explicit use of exporters and providers. Changes are noted in the `opentelemetry_otlp` [v0.27.0 update](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-otlp/CHANGELOG.md#0270) , particularly with the changes from `https://github.com/open-telemetry/opentelemetry-rust/pull/2221`. The protocol (grpc vs. http) is only needed to determine the exporter, and the providers are built after with the exporter.
- The [Deprecated Config public methods](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/CHANGELOG.md#0271) will show up as warnings so these were also updated, where `resource` and `sampler` move to the providers.

Closes: #274

### DIFF
--- a/src/utils/trace.rs
+++ b/src/utils/trace.rs
@@ -60,16 +60,18 @@ fn init_tracer_provider(
     tracing_config: TracingConfig,
 ) -> Result<Option<opentelemetry_sdk::trace::TracerProvider>, TracingError> {
     if let Some((protocol, endpoint)) = tracing_config.clone().traces {
+        let timeout = Duration::from_secs(3);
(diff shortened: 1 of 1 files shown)


declark1 (approved):
LGTM!

PR #284, opened 2025-01-23
:wrench: Assume http2 for grpc clients

evaline-ju:
There have been some reports of `HTTP/2 was not negotiated` along with `tls handshake eof` when the grpc client is used for either tokenization or generation. This configuration appears to mitigate those errors.

More context around avoiding ALPN negotiation is in the referenced issue `https://github.com/hyperium/tonic/issues/1427` and its fix PR `https://github.com/hyperium/tonic/pull/1640`


gkumbhat (approved):
Looks good to me!

PR #133, opened 2024-07-25
Update bug report template

mdevino:
This addresses #132.
Further discussion on what other relevant info might be needed is welcome.

Issue #132, opened 2024-07-25
Update bug report template


mdevino:
## Description

As an orchestrator contributor, I want to add relevant information in the [bug report Platform section](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/fc1ee584ebc8e3282e1a396b3f92e2ddcce86825/.github/ISSUE_TEMPLATE/bug_report.md#platform) , so that I can report bugs in an helpful and straight-forward way.

## Discussion

At this point I can think of two relevant information to add: Rust version and commit hash of the version being tested.
Is there any other relevant info to add to this section?

## Acceptance Criteria

- [ ] [bug_report.md](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/fc1ee584ebc8e3282e1a396b3f92e2ddcce86825/.github/ISSUE_TEMPLATE/bug_report.md) is updated with the expected info.

PR #286, opened 2025-01-24
Update issue template

mdevino:
Reopening as I just noticed #133 was closed.

This addresses #132.
Further discussion on what other relevant info might be needed is welcome.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -19,8 +19,8 @@ Hint: In case you cannot run script due to permissions, try `chmod u+x scripts/c
 <!--
 Example output:
 
-rustc 1.78.0 (9b00956e5 2024-04-29)
-Commit hash: 1892295
+rustc 1.81.0 (eeb90cda1 2024-09-04)
+Orchestrator version: 1.0.0 (279934c)
 -->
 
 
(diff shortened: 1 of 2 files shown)


mdevino:
As requested [here](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/pull/133#pullrequestreview-2248627351), I've updated the script to output as follows:

```
rustc 1.81.0 (eeb90cda1 2024-09-04)
Orchestrator version: 1.0.0 (279934c)
```

In case no tags are associated with the commit HEAD points to:

```
rustc 1.84.0 (9fc6b4312 2025-01-07)
Orchestrator version: No tag associated (1db5d59)
```

Does that work for you, @evaline-ju?

declark1:
Aren't bugs currently being reported through an internal channel? If users are using the released `Dockerfile` in their deployments, it doesn't seem like details other than the release version is needed for a bug report, i.e. they should not need to run a script?

mdevino:
> Aren't bugs primarily reported through an internal channel?
Yes, but since this is an open-source project, I thought we wanted to make it easier for a developer working it to report bugs.

> If users are using the released/versioned Dockerfile and code as-is for their deployments, it doesn't seem like details other than the release version would be needed for a bug report, i.e. they should not need to run a script?
In this case, yes, users would not need to run the script.

> Is the idea to also use this template for internal reporting and mirror bug reports here or have users open bug reports here directly going forward?
I don't think so. I decided to update the bug report template when I was creating my first issues and thought the current template was too generic and not making much sense for our use case.

I'm okay with not merging this PR if we do not see any benefit in updating the template.

PR #288, opened 2025-01-30
output detection for chat completions

resoluteCoder:


src/orchestrator/chat_completions_detection.rs
```diff
@@ -147,31 +172,130 @@ impl Orchestrator {
...
+                    client
+                        .chat_completions(chat_request, c)
+                        .await
+                        .map_err(|error| Error::ChatGenerateRequestFailed {
+                            id: model_id.clone(),
+                            error,
+                        })?;
+
+                if let ChatCompletionsResponse::Unary(ref chat_completion) = chat_completions {
+                    // remove index and update from method to have actual choices
```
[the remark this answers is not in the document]
resoluteCoder: That was an old comment as a reminder for me. I will remove it

src/orchestrator/chat_completions_detection.rs
```diff
@@ -147,31 +172,130 @@ impl Orchestrator {
...
+                                            results: detection_result.results,
+                                        })
+                                        .collect(),
+                                }),
+                                usage: chat_completion.usage.clone(),
+                                warnings: vec![OrchestratorWarning::new(
+                                    DetectionWarningReason::UnsuitableOutput,
+                                    UNSUITABLE_OUTPUT_MESSAGE,
+                                )],
+                                ..Default::default()
```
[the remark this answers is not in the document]
resoluteCoder: Good catch. I went ahead and forwarded all the remaining missing fields from `ChatCompletion`

resoluteCoder:
@evaline-ju this is without any detectors

request
```
{
  "model": "Qwen/Qwen2.5-1.5B-Instruct",
  "guardrail_config": {
    "input": {
      "masks": [],
      "models": {
      }
    }
  },
  "detectors": {
    "input": {
    },
    "output": {
    }
  },
  "max_completion_tokens": 100,
  "n": 3,
  "text_gen_parameters": {
    "max_new_tokens": 25
  },
  "messages": [
    {
      "content": "hello how are you?",
      "role": "system",
      "name": "string"
    }
  ]
}
```

response
```
{
  "id": "chatcmpl-23850f08549d452da9f9a2a5b1812579",
  "object": "chat.completion",
  "created": 1738775062,
  "model": "Qwen/Qwen2.5-1.5B-Instruct",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "Hello! I'm just a computer program, so I don't have feelings. How can I assist you today?"
      },
      "logprobs": null,
      "finish_reason": "stop"
    },
    {
      "index": 1,
      "message": {
        "role": "assistant",
        "content": "Hello! I'm not human, but a computer program. I'm here to help with questions and tasks. How can I assist you today?"
      },
      "logprobs": null,
      "finish_reason": "stop"
    },
    {
      "index": 2,
      "message": {
        "role": "assistant",
        "content": "Hello! I'm a large language model, so I don't have feelings like humans do. But I'm here to help you with anything you need! How can I assist you today?"
      },
      "logprobs": null,
      "finish_reason": "stop"
    }
  ],
  "usage": {
    "prompt_tokens": 13,
    "total_tokens": 106,
    "completion_tokens": 93
  }
}
```

src/orchestrator/chat_completions_detection.rs
```diff
@@ -82,6 +85,25 @@ impl From<&ChatCompletionsRequest> for Vec<ChatMessageInternal> {
...
 
+// Get Vec<ChatMessageInternal> from ChatCompletion
+impl From<&Box<ChatCompletion>> for Vec<ChatMessageInternal> {
+    fn from(value: &Box<ChatCompletion>) -> Self {
+        value
+            .choices
+            .iter()
+            .enumerate()
+            .map(|(index, choice)| ChatMessageInternal {
+                message_index: index,
```
gkumbhat, line 96: Not sure if choices are guaranteed to be in order, so should we use `choice.index` directly here instead of going through the `enumerate` path?

src/orchestrator/chat_completions_detection.rs
```diff
@@ -147,31 +173,129 @@ impl Orchestrator {
...
                     .clients
                     .get_as::<OpenAiClient>("chat_generation")
                     .expect("chat_generation client not found");
                 let mut chat_request = task.request;
+                let model_id = chat_request.model.clone();
                 // Remove detectors as chat completion server would reject extra parameter
                 chat_request.detectors = None;
-                client
-                    .chat_completions(chat_request, task.headers)
+                let headers = task.headers.clone();
```
gkumbhat, line 194: this clone can be avoided by directly passing `task.headers` ?

src/orchestrator/chat_completions_detection.rs
```diff
@@ -147,31 +173,129 @@ impl Orchestrator {
...
-                        id: model_id,
+                        id: model_id.clone(),
                         error,
-                    })
+                    })?;
+
+                if let ChatCompletionsResponse::Unary(ref chat_completion) = chat_completions {
+                    let choices = Vec::<ChatMessageInternal>::from(chat_completion);
+
+                    let output_detections = match detectors.output {
```
gkumbhat, line 206: nit: since this function has gotten quite big, can we split out all of this logic (fetching the output detection part) into separate function ?

src/orchestrator/chat_completions_detection.rs
```diff
@@ -147,31 +173,129 @@ impl Orchestrator {
...
+                                    // sort detection by starting span, if span is not present then move to the end of the message
+                                    detection.results.sort_by_key(|r| match r {
+                                        GuardrailDetection::ContentAnalysisResponse(value) => {
+                                            value.start
+                                        }
+                                        _ => last_idx,
+                                    });
+                                    detection
+                                })
+                                .collect::<Vec<_>>();
```
gkumbhat, line 265: nit: since this is basically same as the processing done at input time, can we move this logic into a separate function and then reuse in both input and output time?

gkumbhat (changes requested):
Thanks @resoluteCoder for the contributions.. I left a few questions and comments

src/orchestrator/chat_completions_detection.rs
resoluteCoder: I get an error if I try to do that since I'm using `task.headers` inside the closure on line `212`.

```
borrow of moved value: `task.headers`
move occurs because `task.headers` has type `HeaderMap`, which does not implement the `Copy`
chat_completions_detection.rs(196, 53): value moved here
chat_completions_detection.rs(212, 51): borrow occurs due to use in closure
borrow of moved value: `task.headers`
move occurs because `task.headers` has type `http::HeaderMap`, which does not implement the `Copy` 
chat_completions_detection.rs(196, 53): value moved here
chat_completions_detection.rs(212, 51): borrow occurs due to use in closure
```

A lot of the clones were me trying to appease the complier :sweat_smile:

src/orchestrator/chat_completions_detection.rs
resoluteCoder: I agree that they are similar and we could break that into a `sort_detections` function. However when I was doing just that I remembered in the input portion I returned the `InputDetectionResult` instead of the `DetectionResult` effectively `casting` it to the appropriate type.

If I do that with the output section I can remove  where I do the conversion on line 281-287.

or of course I can remove the "casting" in those and take out the similar functionality and then loop through and "cast" them to their appropriate type.

I'm good with whichever. Regardless I did kind of mix and match those solutions :sweat_smile: 

Here is what it would look like

Option 1 - more iterating with concise functionality
```rust
fn sort_detections(mut detections: Vec<DetectionResult>) -> Vec<DetectionResult> {
    detections.sort_by_key(|value| value.index);

    detections
        .into_iter()
        .map(|mut detection| {
            let last_idx = detection.results.len();
            // sort detection by starting span, if span is not present then move to the end of the message
            detection.results.sort_by_key(|r| match r {
                GuardrailDetection::ContentAnalysisResponse(value) => value.start,
                _ => last_idx,
            });
            detection
        })
        .collect::<Vec<_>>()
}
```

```rust
    detections: Some(ChatDetections {
                        input: detections
                            .into_iter()
                            .map(|detection_result| InputDetectionResult {
                                message_index: detection_result.index,
                                results: detection_result.results,
                            })
                            .collect(),
                        output: vec![],
                    }),
```

Option 2 - less iterating, repetitive functionality
```rust
                let detections = input_detections
                    .into_iter()
                    .map(|mut detection| {
                        let last_idx = detection.results.len();
                        // sort detection by starting span, if span is not present then move to the end of the message
                        detection.results.sort_by_key(|r| match r {
                            GuardrailDetection::ContentAnalysisResponse(value) => value.start,
                            _ => last_idx,
                        });
                        // detection
                        InputDetectionResult {
                            message_index: detection.index,
                            results: detection.results,
                        }
                    })
                    .collect::<Vec<_>>();
```

src/orchestrator/chat_completions_detection.rs
resoluteCoder: Sounds good. I will break out the output detections into its own function.

resoluteCoder:
@gkumbhat went ahead and choose option 1 and separated the output detections and sort detection logic in their own function

declark1 (commented):
While reviewing this, I ended up having a lot of nit comments around cleanups and restructuring the processing steps, including:

- simplify function names, e.g.
  - message_detections() -> detect()
  - detector_chunk_task() -> chunk()
- chunk() should be refactored to take a list of chunker_ids and messages, returning a map of chunker_id->chunks, rather than a map of detectors as multiple detectors can use the same chunker
- detect() should only perform detection
  - preprocessing (only applicable to input detections) should be moved out of detect()
  - pass chunks to detect() instead of messages
  - return detections sorted from detect() rather than subsequently calling sort_detections()
- drop handle_output_detections() and process in match block for consistency with input detections
- mutate ChatCompletion instead of building a new object
- misc other things

I experimented with these changes and more on a fork [here](https://github.com/declark1/fms-guardrails-orchestrator/blob/chat-output-detection-tweaks/src/orchestrator/chat_completions_detection.rs). Since a lot of my suggested changes are applicable to both input/output detection, I won't request the changes here. We also have a broader refactor in the works to cleanup and reuse code across handlers, which will replace some of this code with common code. I'll open a separate issue/PR for items that don't overlap with the refactor. This review was still very helpful for designing common code.

gkumbhat (commented):
Thanks @resoluteCoder for making the changes. Agree with some of the suggestions that @declark1 made. But if we can make those in separate PR, thats fine too.

gkumbhat (approved):

PR #298, opened 2025-02-07
Integration tests for detection content endpoint

mdevino:
With the release of Mocktail, we could leverage its capabilities to create our integration testing suite for orchestrator. This PR contains two tests for the `/api/v2/text/detection/content` to showcase Mocktail's unary mocking capabilties. The test cases are:

1. Test request with a detector associated with whole_doc_chunker.
2. Test request with a detector associated with a sentence chunker.

I also added a test for a chunker unary call.


mdevino:
Question: the compiler is marking code in `tests/common/mod.rs` as unused, even though it is used. Is there any way to fix that besides adding the `#[allow(dead_code)]` macro?

declark1 (commented):
Why are changes in `server.rs` needed?

tests/common/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+///
+pub fn ensure_global_rustls_state() {
+    let _ = ring::default_provider().install_default();
+}
+
+/// Starts mock servers and adds them to orchestrator configuration.
+pub async fn create_orchestrator_shared_state(
+    detectors: Vec<HttpMockServer>,
+    chunkers: Vec<(&str, MockChunkersServiceServer)>,
+) -> Arc<ServerState> {
```
declark1, line 45: What about the other client types?

tests/common/mod.rs
```diff
@@ -0,0 +1,76 @@
...
+    let mut config = OrchestratorConfig::load(CONFIG_FILE_PATH).await.unwrap();
+
+    for detector_mock_server in detectors {
+        let _ = detector_mock_server.start().await;
```
declark1, line 49: This would ignore errors during startup. We should use `mock_server.start().await?` and return `Result`

tests/common/mod.rs
mdevino: I didn't create any tests that use different clients yet, but I agree it's a problem. Do Mocktail gRPC clients implement a common trait? If so, would it be possible to make this change?

tests/detection_content.rs
```diff
@@ -0,0 +1,221 @@
...
+use tracing_test::traced_test;
+
+mod common;
+
+// Constants
+const ENDPOINT_ORCHESTRATOR: &str = "/api/v2/text/detection/content";
+const ENDPOINT_DETECTOR: &str = "/api/v1/text/contents";
+
+const DETECTOR_NAME_ANGLE_BRACKETS_WHOLE_DOC: &str = "angle_brackets_detector_whole_doc";
+const DETECTOR_NAME_ANGLE_BRACKETS_SENTENCE: &str = "angle_brackets_detector_sentence";
```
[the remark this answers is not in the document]
mdevino: I initially implemented this as a detector that detects the word "word", but figured the angle brackets would make it easier to visualise the detection. Would you think this detector would be easier to read?

mdevino:
> Why are changes in server.rs needed?
Because `axum_test.TestClient` needs a Router to be constructed.

tests/common/mod.rs
declark1: No, they don't. The trait is the generated gRPC service, which is why I don't think this approach is a good idea as it will end up being more complicated.

tests/common/mod.rs
mdevino: @declark1 Since this function won't exist anymore, is it okay to use this approach on test functions, or would it be cleaner to return a `Result`? I ask because tests are going to fail regardless.

tests/common/mod.rs
mdevino: Actually, now that I think about it, there would be no other client types for chunkers, right? The only one we have is listed [here](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/c202774b81f745e9df8f919265d2d031219cf23a/protos/caikit_runtime_Chunkers.proto#L30).

mdevino:
> Question: the compiler is marking code in `tests/common/mod.rs` as unused, even though it is used. Is there any way to fix that besides adding the `#[allow(dead_code)]` macro?

Reminder, in case the previous question got lost in the conversations.

tests/common/mod.rs
mdevino: Also, this change is making clippy fail, as shown [here](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/actions/runs/13241695972/job/36958317841#step:7:554).
Would you know how to fix this without using the `allow` macro? I came across [this link](https://rust-lang.github.io/rust-clippy/master/index.html#option_map_unit_fn), but my understanding is that the suggestion provided will not work if the `Option` value is `None`.

tests/common/mod.rs
mdevino: Looks like #[tokio::test] doesn't allow returning `Result`, so `unwrap()` is needed at the test level.

src/config.rs
```diff
@@ -355,6 +355,28 @@ impl OrchestratorConfig {
...
+
+    pub fn set_chunker_port(&mut self, name: &str, port: u16) {
+        self.chunkers
+            .as_mut()
+            .map(|chunkers| chunkers.get_mut(name).unwrap().service.port = Some(port));
+    }
+
+    pub fn set_detector_port(&mut self, name: &str, port: u16) {
+        self.detectors.get_mut(name).unwrap().service.port = Some(port);
+    }
```
declark1, line 379: Re: the clippy `option_map_unit_fn` lint, update the methods to the following. 

Technically, returning `Result<(), Error>` would be more appropriate if the client config doesn't exist, but then we would need to add a new `Error` variant for this scenario. Since these methods are just used for integration testing, I think this behavior is fine: if a client config doesn't exist, the port isn't updated. You shouldn't be updating something that doesn't exist anyway.

```suggestion
    pub fn set_generation_port(&mut self, port: u16) {
        if let Some(generation) = self.generation.as_mut() {
            generation.service.port = Some(port);
        }
    }

    pub fn set_chat_generation_port(&mut self, port: u16) {
        if let Some(chat_generation) = self.chat_generation.as_mut() {
            chat_generation.service.port = Some(port);
        }
    }

    pub fn set_chunker_port(&mut self, name: &str, port: u16) {
        if let Some(chunkers) = self.chunkers.as_mut() {
            if let Some(chunker) = chunkers.get_mut(name) {
                chunker.service.port = Some(port);
            }
        }
    }

    pub fn set_detector_port(&mut self, name: &str, port: u16) {
        if let Some(detector) = self.detectors.get_mut(name) {
            detector.service.port = Some(port);
        }
    }
```

tests/common/mod.rs
declark1: Not accurate, you can definitely return `Result` when `#[tokio::test]` is used; I have definitely done this before.

tests/common/mod.rs
declark1: > @declark1 I've implemented the change you suggested, but I really think having a separate helper function to deal with starting servers/tweaking orchestrator config makes more sense. The rationale behind my thinking is that if I'm writing a test I would rather focus on the test logic rather than having to think about what I need to initialize.

Creating a helper function to do this isn't trivial when you are dealing with multiple types of clients, including those generated at compile time for gRPC. I would argue that developing a hacky solution to do this isn't worth the effort required just because you don't want to call `server.start()` and set the port on the config.

mdevino:
> I'm unaware of any other workaround. I would expect the compiler to detect that the common code is used, but apparently not.

Looks like this happen because each test file is compiled as a separate module. If any of the pub code in common.rs is not used by any of the test files, it generates this warning ([source](https://github.com/rust-lang/rust/issues/46379)).

I tried implementing the workaround suggested [here](https://matklad.github.io/2021/02/27/delete-cargo-integration-tests.html), but I couldn't get it to work, so I just added the `allow` macros.

tests/common/mod.rs
mdevino: As we discussed offline, I'm maintaining the helper function. Since the macros to generate Mocktail gRPC servers are being called in `test/common/mod.rs`, all test files would be using the same generated type `MockChunkersServiceServer`.

Thanks for the discussion!

mdevino:
@declark1 I think all changes requested have been addressed, except [this one](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/pull/298#discussion_r1946883223). Is there a detector concept you think would be easier to illustrate than angle brackets?

PR #306, opened 2025-02-11
:wrench: Make grpc dns probe interval configurable per client

gkumbhat:
### Changes

- Make grpc probe configurable per client
- Fix default connect timeout


gkumbhat:
@evaline-ju true, but thought of being more specific about it, in case there is a different lingo used in other language.

gkumbhat:
@evaline-ju changed it, looking at the nomenclature ginepro uses.

Issue #289, opened 2025-01-31
Integration tests for `/text/detection/content`


evaline-ju:
## Description

As an orchestrator developer, I want to have integration tests for the [standalone content detection endpoint](https://foundation-model-stack.github.io/fms-guardrails-orchestrator/?urls.primaryName=Orchestrator+API#/Task%20-%20Detection/api_v2_detection_text_content_unary_handler), so that I can make changes and make sure I am not breaking existing logic.

## Acceptance Criteria

- [ ] Integration tests exist
- [ ] Overlapping tests that required mocking clients removed, if existent
- [ ] Appropriate use of [mocktail](https://github.com/IBM/mocktail) as necessary

mdevino:
The list of tests I can think of are:

* [x] 1. detector (whole_doc chunker), no detection -> should return no detection
* [x] 2. detector (sentence chunker), no detection -> should return generated text
* [x] 3. detector (whole_doc chunker), with detection -> should return detections
* [x] 4. detector (sentence chunker), with detection -> should return detections
* [x] 5. detector returning 503 -> should return 503
* [x] 6. detector returning 404 -> should return 404
* [x] 7. detector returning 500 -> should return 500
* [x] 8. detector returning message not compliant with API -> should return 500
* [x] 9. chunker return error -> should return 500
* [x] 10. orchestrator receives incorrect request (extra fields) -> should return 422
* [x] 11. orchestrator receives incorrect request (missing detectors) -> should return 422
* [x] 12. orchestrator receives incorrect request (missing content) -> should return 422
* [x] 12. orchestrator receives incorrect request (detectors is an empty map) -> should return 422

PR #314, opened 2025-02-25
Integration tests for `/api/v2/text/detection/content`

mdevino:
This addresses #289 with the test cases listed [here](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/issues/289#issuecomment-2679801819).


declark1 (changes requested):
For this endpoint, I think we only need 4 test scenarios:
- no_detections
- detections
- client_error
- orchestrator_validation_error

I don’t think we need to test different chunker types here for a unary endpoint, but if you want to, combine them into a single test.

Also, let’s be consistent with the naming conventions we established. It should be “no_detections” (plural) and the test function name should omit the “test_” prefix (it’s redundant for integration tests since everything is a test). I'll create a README for integration tests folder with the naming conventions, etc. for future reference.

mdevino:
@declark1 Changes addressed.

declark1 (approved):
LGTM

PR #333, opened 2025-03-07
Refactor TestOrchestratorServerBuilder and TestOrchestratorServer to avoid mock server ownership

declark1:
As mocktail now supports mutating mocks on the server at runtime, e.g.

```rust
server.mocks().clear();
server.mocks().insert(...);
server.mocks().entry(...);
server.mocks().remove(...);
```

The initial approach to passing ownership of the `MockServer`s to the `TestOrchestratorServerBuilder` and subsequently `TestOrchestratorServer`, makes calling these methods tricky for integration tests. This PR introduces changes to only pass references to the builder. Existing tests have been updated accordingly.

### DIFF
--- a/tests/common/orchestrator.rs
+++ b/tests/common/orchestrator.rs
@@ -16,7 +16,6 @@
 */
 
 use std::{
-    collections::HashMap,
     marker::PhantomData,
     net::{IpAddr, Ipv4Addr, SocketAddr},
     pin::Pin,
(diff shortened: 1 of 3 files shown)


mdevino (approved):
Looks good to me!
I'm still confused about how `<'a>` is enough for Rust to understand the reference will live long enough, but I've ran a couple of tests locally and I can see that 1) tests continue passing  and 2) mock server variables can be edited after the orchestrator test server has been initialized.

gkumbhat (approved):
Looks good to me!

PR #334, opened 2025-03-07
:memo: ADR and API updates for additional detector fields

evaline-ju:
- ADR with decision of using `metadata`
- Consequent API updates for other orchestrator and detector APIs [the detector APIs currently mostly have the same response structure, but separate objects have been maintained in case of divergence, so the `metadata` update has been made separately to each object. This could be refactored in the future.]
- Updates status of previous ADRs

Relates to supporting: https://github.com/foundation-model-stack/vllm-detector-adapter/issues/32

### DIFF
--- a/docs/architecture/adrs/011-detector-fields.md
+++ b/docs/architecture/adrs/011-detector-fields.md
@@ -21,7 +21,7 @@ Any placement of additional detector fields should account for the two types of 
 - Knowledge of future model plans is being restricted here, so only a few examples are given with already released model functionalities.
 
 ## Decisions
-- We will add a new high-level field of `metadata` to account for additional model information from detectors. This field will provide a dictionary with string keys and arbitrary values, so that values are not constrained to particular types like strings or floats. This will enable flexibility and is how APIs like [Llama Stack](https://github.com/meta-llama/llama-stack) provides additional information, whether on datasets or models.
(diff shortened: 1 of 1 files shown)


docs/architecture/adrs/011-detector-fields.md
```diff
@@ -0,0 +1,143 @@
...
+2. Other endpoints return a list of detections, without spans.
+
+Any placement of additional detector fields should account for the two types of responses.
+
+## Assumptions
+- This ADR considers fields that apply on each particular detection or decision made by the detector model.
+- Knowledge of future model plans is being restricted here, so only a few examples are given with already released model functionalities.
+
+## Decisions
+- We will add a new high-level field of `metadata` to account for additional model information from detectors. This field will provide a dictionary with string keys and arbitrary values, so that values are not constrained to particular types like strings or floats. This will enable flexibility and is how APIs like [Llama Stack](https://github.com/meta-llama/llama-stack) provides additional information, whether on datasets or models.
```
gkumbhat, line 24: nit: 
```suggestion
- We will add a new high-level field of `metadata` to account for additional information from detector models. This field will provide a dictionary with string keys and arbitrary values, so that values are not constrained to particular types like strings or floats. This will enable flexibility and is how APIs like [Llama Stack](https://github.com/meta-llama/llama-stack) provides additional information, whether on datasets or models.
```

Since its not really additional model information, but information about the output that model is providing.

gkumbhat (approved):
Small suggestion, but otherwise looks good!

PR #336, opened 2025-03-12
reverted response_stream processing and message json conversion for stream content detection

swith004:
This PR reverts the response stream handling for the stream content detection to remove the bug introduced during validation for the detectors field when jsonlines was used as response.

The bug was seen with the following request: 
```
curl -X 'POST' \
  'http://localhost:8033/api/v2/text/detection/stream-content' \
  -H 'accept: application/x-ndjson' \
  -H 'Content-Type: application/x-ndjson' \
  -d '{"content": "hello"}'
curl: (18) transfer closed with outstanding read data remaining
```

The expected response is the following: 
`{'code': 422, 'details': '`detectors` is required for the first message'}`

The problem is how processing of the errors occurs in jsonlines which requires prior error mapping and serialization. Going back to the original code will remove the regressive behavior.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -38,14 +38,14 @@ use axum::{
 };
 use axum_extra::{extract::WithRejection, json_lines::JsonLines};
 use futures::{
-    stream::{self, BoxStream, TryStreamExt},
+    stream::{self, BoxStream},
     Stream, StreamExt,
 };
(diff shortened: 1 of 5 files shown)


declark1 (approved):
LGTM, thanks for diving deep into this issue.

Issue #311, opened 2025-02-24
Integration tests for output detections in `/api/v1/task/server-streaming-classification-with-text-generation`  [NLP]


mdevino:
This is a follow-up of #299.

## Description

As an orchestrator developer, I want to have a robust test suite, so that I can refactor the code with a lower chance of introducing bugs.

## Discussion

The intent of this story is to agree and implement on a basic integration test suite for the [/api/v1/task/server-streaming-classification-with-text-generation endpoint](https://foundation-model-stack.github.io/fms-guardrails-orchestrator/?urls.primaryName=Orchestrator+API#/Task%20-%20Text%20Generation%2C%20with%20detection/guardrails_server_stream_handler_api_v1_text_server_streaming_classification_with_text_generation_post). We can use [Mocktail](https://github.com/IBM/mocktail) for tests.

The list of tests I can think of are:

* [x] 1. output detector (sentence chunker), no detection -> should return generated text
* [x] 2. output detector (sentence chunker), with detection -> should return generated text AND output detection
* [x] 3. output detector returning 503 -> should return 503
* [x] 4. output detector returning 404 -> should return 404
* [x] 5. output detector returning 500 -> should return 500
* [ ] 6. output detector returning message not compliant with API -> should return 500
* [ ] 7. output chunker return error -> should return 500

## Acceptance Criteria

- [ ] Common agreement on the initial list of test cases.
- [ ] Implementation of the agreed test cases.

PR #342, opened 2025-03-17
Output streaming integration tests

mdevino:
Addresses #311.


tests/classification_with_text_gen.rs
gkumbhat, line 1: QQ: why was this name changed ? Since we are adding streaming tests in separate files, it seems intuitive to have a file with `unary` ? Also the name here does not include `nlp` whereas it does for the streaming one.. Can we make these names align?

declark1: The idea is for the names to match the handler files coming in the refactor, which align with the names of the endpoints. "Unary" is not part of the endpoint name.

gkumbhat: Ah I see what you mean. Yea, adding `nlp` will be better, since we _might_ want to add test for tgis as well in near future

declark1: Although, I don't see why nlp and tgis tests can't be in the same file since they are for the same endpoint. It would probably be better to indicate this in the test name rather than different files?

tests/streaming_classification_with_gen_nlp.rs
```diff
@@ -648,3 +659,779 @@ async fn orchestrator_validation_error() -> Result<(), anyhow::Error> {
...
+
+    assert_eq!(
+        messages[1].generated_text,
+        Some(" What about <you>?".into())
+    );
+    assert_eq!(
+        messages[1].token_classification_results.output,
+        Some(vec![TokenClassificationResult {
+            start: 13,
+            end: 16,
```
gkumbhat, line 1131: nit: these seems wrong for final output ?  since the end span for 2nd sentence is 31 according to here: https://github.com/foundation-model-stack/fms-guardrails-orchestrator/pull/342/files#diff-49489e07e62a8b7afb367938f7f75e7e8a9480135d0006d638c0b590cc3bc3c2R979


might get a bit confusing

tests/classification_with_text_gen.rs
mdevino: What I really wanted was to make `streaming_classification_with_gen` a folder, then create files `nlp.rs` and `tgis.rs` inside, but I couldn't get tests inside the folder to be discovered.

I thought we wanted to split files to prevent files from being too big (though they will be, anyway), but I don't have strong opinions on combining tests in the same file (maybe separate tests inside a `mod` in this approach?).

tests/streaming_classification_with_gen_nlp.rs
```diff
@@ -648,3 +659,779 @@ async fn orchestrator_validation_error() -> Result<(), anyhow::Error> {
...
+                input_start_index: 0,
+                input_end_index: 0,
+            },
+        ]);
+    });
+
+    // Add output detection mock
+    let expected_detector_error = DetectorError {
+        code: 500,
+        message: "The detector service is overloaded.".into(),
```
gkumbhat, line 1311: nit: overloaded is generally 503 or 429, depending on ratelimting or server error..   Might want to use any other error message to avoid confusion.

tests/classification_with_text_gen.rs
declark1: Edited my comment above after thinking about this more. IMO, we don't need need separate sets of integration tests for this, we can just use the nlp or tgis variant for all of them. We can have separate tests to ensure parity between NLP/TGIS variants of the GenerationClient.

gkumbhat: since the clients have pretty different APIs (or can have), the parity test would end up looking a lot more like the integration test..


I like the idea of having `streaming_classification_with_gen` as a folder.  Thought @declark1 ?

mdevino: Sure. If we want to go down this route, I'll remove the `_nlp` at the end of the files.

tests/streaming_classification_with_gen_nlp.rs
mdevino: @gkumbhat Since I renamed the file I lost track of this comment. Could you make it again in the renamed file?

tests/streaming_classification_with_gen.rs
```diff
@@ -0,0 +1,1437 @@
...
+            word: "you".into(),
+            entity: "has_angle_brackets".into(),
+            entity_group: "angle_brackets".into(),
+            detector_id: Some(angle_brackets_detector.into()),
+            score: 1.0,
+            token_count: None
+        }])
+    );
+    assert_eq!(messages[1].start_index, Some(13));
+    assert_eq!(messages[1].processed_index, Some(31));
```
declark1, line 1141: Instead of all these assertions, wouldn't it be clearer to just `assert_eq!(messages, expected_messages)` ?

declark1 (commented):
Mostly looks good, but I left one comment/question inline

tests/classification_with_text_gen.rs
declark1: I still don't see why we need separate tgis/nlp integration tests, and thus a folder for them, but it's up to you guys. The differences between these clients aren't relevant to these tests and they are wrapped in `GenerationClient` anyway.

gkumbhat: @declark1  I agree with what you are saying @declark1 .  For me it boils down to, how we will be testing TGIS client (and future such clients).

So whichever method is quickest to make sure functionality is working with TGIS as well.

tests/streaming_classification_with_gen.rs
mdevino: I've made the change. The code might look cleaner, but it might be a little harder to spot what is the difference when looking at logs if the test fails.

gkumbhat (approved):
Looks good to me!

mdevino:
Hi. Could we merge this one, so I can rebase and open the next batch of tests?

Issue #343, opened 2025-03-19
CI checks might not be working as expected


mdevino:
This repo has a GitHub actions workflow that runs on opened PRs. There were a few instances in which checks passed when they shouldn't (a recent example being formatting checks passing in #340).

I have also noticed that sometimes checks take 6\~7 minutes, while sometimes it only takes around 12\~20 seconds. This issue might be related to the caching mechanism used, but further investigation is needed.

declark1:
Fixed with #346

Issue #308, opened 2025-02-12
Skip CI for Draft PRs


declark1:
WIP branches typically have high commit frequency and running CI jobs for each commit is a waste of resources. Additionally, incomplete code is likely to trigger clippy lint failures such as dead code, unused variables, etc. I was thinking it would be nice to skip Test + Lint CI jobs for Draft PRs, only running them once a PR is in a Ready for Review state. 

Thoughts?

I believe this can be accomplished with the following condition:
```
if: github.event.pull_request.draft == false
```

mdevino:
I agree. CI jobs shouldn't run on draft PRs. If for some reason one would like to run the CI jobs, they could mark the PR as ready for review, then change back later to draft if needed.

declark1:
Completed with #346

PR #346, opened 2025-03-22
Update CI checks

mdevino:
This PR addresses #343 and #308.

A couple of notes about our current CI process:

* Checks are applied only to the main code (under `/src`).
* We have configuration in `rustfmt.toml` that is only available in nightly builds, while we're using `fmt` of the stable channel.
* There seems to be unnecessary docker-related dependencies in our GitHub actions (we do not distribute container images at this point).
* Our current CI also runs on commits to main, so they actually run again once we merge a PR.

To address these topics, I have:

1. Created a new GitHub Actions workflow that does not use the docker-related dependencies, and runs on all Rust targets (tests included). It also uses `fmt` from the nightly channel. There is a step to install the same version of `protoc` we're using on the Dockerfile.
2. Updated pre-commit hooks to use fmt from the nightly channel.
3. Applied fmt nightly to all files in the repo (hence the import changes on .rs files)

What I've tried:
1. Caching seems to make a difference, so I've kept a caching mechanism in the workflow.
2. Setting `CARGO_INCREMENTAL=0` seems to make things slower (most likely due to caching making things faster), so I did not add it.

Once this PR is merged, it's necessary to run `pre-commit autoupdate` locally after pulling, so hooks are updated.
In case the `fmt-nightly` hook fails because the nightly toolchain is not installed, a command will be provided to install it (see the image below for what I see on Linux/WSL).

image


.github/workflows/test.yml
```diff
@@ -1,93 +1,43 @@
-name: Test and Lint Orchestrator
+name: Rust
 
 on:
-  push:
```
gkumbhat, line 4: We do want to run CI whenever new changes gets pushed to `main` branch, this would technically always happen via pull requests, but this gives a baseline and sanity checking.

.pre-commit-config.yaml
```diff
@@ -1,7 +1,7 @@
 repos:
-  - repo: https://github.com/doublify/pre-commit-rust
-    rev: v1.0
+  - repo: https://github.com/mdevino/pre-commit-rust
```
gkumbhat, line 2: If we need to overwrite the `.pre-commit-hooks.yaml` file, can that file be committed to this repo itself ?

gkumbhat (commented):
Nice work @mdevino . Left some comments and suggestions

.github/workflows/test.yml
mdevino: I could add it back, but could you elaborate on why this is needed?

The way I see it, the only use for running checks on main is if we push directly to it.

mdevino: I guess I found a reason. We don't enforce closing PRs by rebasing, so merging into main does have a chance to break things. I've added it back.

.github/workflows/test.yml
```diff
@@ -1,93 +1,43 @@
-name: Test and Lint Orchestrator
+name: Rust
```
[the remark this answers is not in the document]
mdevino: Done.

.pre-commit-config.yaml
declark1: I think @gkumbhat is wondering why you're pulling this from your fork just to override the `.pre-commit-hooks.yaml`, rather than adding that file to this repo.

mdevino: Ah that. Not sure how, but I missed the filename in Gaurav's message when I read it this morning.
Technically I could, but I'm wondering if this is a good idea. The revision is supposed to be a fixed reference. Do we want to start creating tags to version the hooks file?
Maybe we should push it to a separate repo under the org.

declark1: I'm not very familiar with pre-commit hooks, so I don't know the best practice. If that's the case, this is probably fine then since creating a new repo just for this in the IBM org probably isn't practical.

mdevino: @declark1 Do you have the permissions required to fork https://github.com/doublify/pre-commit-rust into this organization? I tried changing the ownership of my fork, but I don't have the sufficient permissions.

declark1: It is a process to create new repos in the IBM org. I don't see why we need to do this?

mdevino: I just found out about [local hooks](https://pre-commit.com/#repository-local-hooks), so we don't need to fork it. I've updated the configuration accordingly. Thanks for pointing that out, @declark1!

declark1 (approved):
Looks good, thanks!

Issue #349, opened 2025-03-26
Error handling for `/stream-content` endpoint


mdevino:
## Describe the bug

Looks like there are a few error scenarios orchestrator is not handling correctly on the `/stream-content` endpoint. Particularly, 1) when a request is sent without the `content` field and 2) when there is an extra argument in the request. Orchestrator is returning 500 when it should be a 422.

I'm sharing  two tests below illustrating these scenarios.

```rust
#[test(tokio::test)]
async fn request_with_extra_args() -> Result<(), anyhow::Error> {
    let detector_name = DETECTOR_NAME_ANGLE_BRACKETS_SENTENCE;

    // Run test orchestrator server
    let orchestrator_server = TestOrchestratorServer::builder()
        .config_path(ORCHESTRATOR_CONFIG_FILE_PATH)
        .build()
        .await?;

    // assert extra argument on request
    let response = orchestrator_server
        .post(ORCHESTRATOR_STREAM_CONTENT_DETECTION_ENDPOINT)
        .header("content-type", "application/x-ndjson")
        .body(reqwest::Body::wrap_stream(json_lines_stream([json!( {
            "detectors": {detector_name: {}},
            "content": "Hi there!",
            "extra_arg": true
        })])))
        .send()
        .await?;
    let mut messages = Vec::<OrchestratorError>::with_capacity(1);
    let mut stream = response.bytes_stream();
    while let Some(Ok(msg)) = stream.next().await {
        debug!("recv: {msg:?}");
        messages.push(serde_json::from_slice(&msg[..]).unwrap());
    }
    assert_eq!(messages.len(), 1);
    assert_eq!(messages[0].code, 422);
    assert!(messages[0].details.starts_with("unknown field `extra_arg`"));

    Ok(())
}

#[test(tokio::test)]
async fn request_missing_content() -> Result<(), anyhow::Error> {
    let detector_name = DETECTOR_NAME_ANGLE_BRACKETS_SENTENCE;

    // Run test orchestrator server
    let orchestrator_server = TestOrchestratorServer::builder()
        .config_path(ORCHESTRATOR_CONFIG_FILE_PATH)
        .build()
        .await?;

    // assert missing `content` on first frame
    let response = orchestrator_server
        .post(ORCHESTRATOR_STREAM_CONTENT_DETECTION_ENDPOINT)
        .header("content-type", "application/x-ndjson")
[...]

Issue #322, opened 2025-02-26
Integration tests for `/api/v2/text/detection/stream-content`


mdevino:
## Description

As an orchestrator developer, I want to have a robust test suite, so that I can refactor the code with a lower chance of introducing bugs.

## Discussion

The intent of this story is to agree and implement on a basic integration test suite for the [/api/v2/text/detection/stream-content endpoint](https://foundation-model-stack.github.io/fms-guardrails-orchestrator/?urls.primaryName=Orchestrator+API#/Task%20-%20Detection/api_v2_detection_text_content_bidi_stream_handler). We can use [Mocktail](https://github.com/IBM/mocktail) for tests.

The list of tests I can think of are:

* [ ] 1. no detectors, just generation -> should return generated text
* [ ] 2. detector (sentence chunker), no detection -> should return generated text
* [ ] 3. detector (sentence chunker), with detection -> should return generated text AND detection
* [ ] 4. detector returning 503 -> should return 503
* [ ] 5. detector returning 404 -> should return 404
* [ ] 6. detector returning 500 -> should return 500
* [ ] 7. detector returning message not compliant with API -> should return 500
* [ ] 8. chunker return error -> should return 500
* [ ] 9. generation server returning error -> should return 500
* [ ] 10. orchestrator receives incorrect request -> should return 422

## Acceptance Criteria

- [ ] Common agreement on the initial list of test cases.
- [ ] Implementation of the agreed test cases.

PR #350, opened 2025-03-26
Stream content tests and error handling fix

mdevino:
This addresses #322 and #349.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -455,10 +455,12 @@ async fn stream_content_detection(
 
     // Create input stream
     let input_stream = json_lines
-        .map(|result| {
-            let message = result.unwrap();
-            message.validate()?;
-            Ok(message)
+        .map(|result| match result {
+            Ok(message) => {
+                message.validate()?;
+                Ok(message)
+            }
+            Err(error) => Err(orchestrator::errors::Error::Validation(error.to_string())),
         })
         .boxed();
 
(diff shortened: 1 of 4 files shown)


mdevino:
Converting this back to draft while I'm still investigating #349. Test code shouldn't change though, in case anyone would like to start reviewing.

declark1:
The tests in the issue appear to pass here. Did your fix here not resolve it?

mdevino:
> The tests in the issue appear to pass here. Did your fix here not resolve it?

I just figured it out. These tests are correct, it's just that there was an issue with the functional tests we run internally. I wanted to make sure these changes were not breaking anything.

PR ready for review again.

mdevino:
There's an unrelated test failing because a subset of a json response is being serialized in a different order. Working on it.
image

declark1:
> There's an unrelated test failing because a subset of a json response is being serialized in a different order, making the mock match fail. Working on it.

Hmm. Ordering of keys in a map should not impact equality. What is the specific type failing?

declark1 (commented):
Can we add a multi-detector test here like we have for the streaming classification with gen output detections? Otherwise, we aren't testing the batching/aggregation. As an efficiency enhancement in the refactored code, batching is skipped entirely if there is only 1 detector applied and we need to test both paths.

mdevino:
> Can we add a multi-detector test here like we have for the streaming classification with gen output detections? Otherwise, we aren't testing the batching/aggregation. As an efficiency enhancement in the refactored code, batching is skipped entirely if there is only 1 detector applied and we need to test both paths.

Good catch. I've edited the existing tests to use two detectors.

declark1 (approved):
LGTM. 

> Can we add a multi-detector test here like we have for the streaming classification with gen output detections? Otherwise, we aren't testing the batching/aggregation. As an efficiency enhancement in the refactored code, batching is skipped entirely if there is only 1 detector applied and we need to test both paths.

Just to clarify this, I actually meant we should have BOTH single and multi-detector tests, to ensure we test non-batching and batching code paths. This is fine for now, but we should create an issue to add single detector tests for the streaming endpoints.

Issue #338, opened 2025-03-14
Add metadata to standalone detection responses


evaline-ju:
## Description

In #334, API updates were documented for both the detector and orchestrator APIs for an additional `metadata` field. As an orchestrator developer, I want to propagate any `metadata` returned from the detector APIs to the end orchestrator user, so that users can also get additional detector model information, if available.

## Discussion

We want to add `metadata` as documented in the API for at least the structs used for [standalone detection orchestrator endpoints](https://foundation-model-stack.github.io/fms-guardrails-orchestrator/?urls.primaryName=Orchestrator+API#/Task%20-%20Detection).

## Acceptance Criteria

- [ ] Unit tests cover new/changed code
- [ ] Examples build against new/changed code
- [ ] READMEs are updated
- [ ] Type of [semantic version](https://semver.org/) change is identified

PR #353, opened 2025-03-27
Add metadata object

gkumbhat:
closes: https://github.com/foundation-model-stack/fms-guardrails-orchestrator/issues/338

### Changes
- Add metadata to `DetectionResult` 
- Add metadata to standalone detection swagger spec examples

### Example Request
```
curl --request POST \
  --url http://localhost:8038/api/v2/text/detection/content \
  --header 'Content-Type: application/json' \
  --data '{
  "detectors": {
    "granite-guardian-content": {
		}
  },
  "content":"Your email is test@ibm.com! Hi, Only the next instance of email will be processed. test@ibm.com. Your SSN is 123-45-6789."
}'
```

### Example Response
```
{
	"detections": [
		{
			"start": 0,
			"end": 121,
			"text": "Your email is test@ibm.com! Hi, Only the next instance of email will be processed. test@ibm.com. Your SSN is 123-45-6789.",
			"detection": "Yes",
			"detection_type": "risk",
			"detector_id": "granite-guardian-content",
			"score": 0.9554123282432556,
			"metadata": {
				"confidence": "High"
			}
		}
	]
}
```

### DIFF
--- a/src/models.rs
+++ b/src/models.rs
@@ -49,9 +49,7 @@ pub struct InfoParams {
 #[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
 pub struct DetectorParams(HashMap<String, serde_json::Value>);
 
-/// Metadata information for each detection
-#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
-pub struct MetadataObj(HashMap<String, serde_json::Value>);
+pub type Metadata = HashMap<String, serde_json::Value>;
 
 impl DetectorParams {
     #[allow(dead_code)]
@@ -925,8 +923,8 @@ pub struct DetectionResult {
     pub evidence: Option<Vec<EvidenceObj>>,
 
     // Optional metadata block
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub metadata: Option<MetadataObj>,
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub metadata: Metadata,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
--- a/src/models.rs
+++ b/src/models.rs
@@ -49,6 +49,10 @@ pub struct InfoParams {
 #[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
 pub struct DetectorParams(HashMap<String, serde_json::Value>);
 
+/// Metadata information for each detection
+#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
+pub struct MetadataObj(HashMap<String, serde_json::Value>);
+
 impl DetectorParams {
     #[allow(dead_code)]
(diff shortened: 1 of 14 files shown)


src/orchestrator/types/detection_batcher/max_processed_index.rs
```diff
@@ -78,7 +78,9 @@ impl DetectionBatcher for MaxProcessedIndexBatcher {
         {
             // We have all detections for the chunk, remove and return it.
             if let Some((chunk, detections)) = self.state.pop_first() {
-                let detections = detections.into_iter().flatten().collect();
+                let mut detections: Detections = detections.into_iter().flatten().collect();
+                // Provide sorted detections within each chunk
+                detections.sort_by_key(|r| r.start);
```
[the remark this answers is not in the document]
gkumbhat: woops.. this should not have been pushed. I'll revert

src/models.rs
```diff
@@ -49,6 +49,10 @@ pub struct InfoParams {
 #[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
 pub struct DetectorParams(HashMap<String, serde_json::Value>);
 
+/// Metadata information for each detection
+#[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
+pub struct MetadataObj(HashMap<String, serde_json::Value>);
```
declark1, line 54: What is the justification for the newtype rather than just using `HashMap<String, serde_json::Value>`? It doesn't appear you are adding any custom validation or functionality to it, so I would just use HashMap.

src/models.rs
```diff
@@ -919,6 +923,10 @@ pub struct DetectionResult {
     // Optional evidence block
     #[serde(skip_serializing_if = "Option::is_none")]
     pub evidence: Option<Vec<EvidenceObj>>,
+
+    // Optional metadata block
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub metadata: Option<MetadataObj>,
```
declark1, line 929: Option isn't needed if you just use `HashMap<String, serde_json::Value>` with `#[serde(default, skip_serializing_if = "HashMap::is_empty")]`

src/models.rs
declark1: For that purpose, type alias is more appropriate

e.g. 
```rust
pub type Metadata = HashMap<String, serde_json::Value>
```

src/models.rs
declark1: It would still work with a type alias though if you want, since it's still the same type

src/models.rs
gkumbhat: ah yes, good idea.. Let me change to that.

gkumbhat:
@declark1 pushed the change. Thanks for the suggestion, that makes much more sense.

declark1 (approved):
LGTM

PR #356, opened 2025-03-28
:construction_worker: Healthcheck command for Dockerfile

evaline-ju:
For `CIS_Docker_v1.5.0 - 4.6` compliance - generally we do not use docker for healthchecking


declark1 (approved):
LGTM

PR #368, opened 2025-04-11
Updated Dockerfile to support s390x

Sanketha-Cr:


declark1:
I'm not familiar with s390x, but would this require compiling the Rust code to a different target as well?

Issue #266, opened 2024-12-24
Drop faux and associated tests


declark1:
## Description

Currently, we use faux for client mocking, which has several limitations for our use case. When we adopt mocktail for server mocking, faux usage should be dropped, with essential tests re-implemented in the integration test stack using server mocking.

## Acceptance Criteria
- [ ] `faux::create` and `faux::methods` attributes dropped from all clients
- [ ] All unit tests utilizing faux have been dropped
- [ ] faux dependency is dropped

declark1:
Completed with #374 

FYI @mdevino , you can add "closes #<issue>", "resolves #<issue>" or other specific keywords in the PR rather than "addresses #" (not a keyword) so that the related issue gets closed automatically when merged.

https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue

mdevino:
> Completed with [#374](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/pull/374)
> 
> FYI [@mdevino](https://github.com/mdevino) , you can add "closes #", "resolves #" or other specific keywords in the PR rather than "addresses #" (not a keyword) so that the related issue gets closed automatically when merged.
> 
> https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue

I'd never guess that. Thanks for letting me know!

PR #374, opened 2025-04-16
Drop faux

mdevino:
Addresses #266. I've simply removed the model info test on `tgis.rs` because the `model_info()` function doesn't seem to be used anywhere, but let me know if I'm missing anything.


declark1 (approved):
Great, thanks!

PR #375, opened 2025-04-16
refactor ChatCompletionsRequest

### LINKED ISSUE #367: Account for additional chat completion parameters for `/chat/completions-detection`
## Description

As an orchestrator user using `chat/completions-detection`, I want to be able to use a variety of parameters that are available in vLLM, like `prompt_logprobs` or `echo`. However, the current [OpenAI client does not include some of those parameters](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/main/src/clients/openai.rs).

## Discussion

We should consider if it's feasible to just pass through extra parameters, or otherwise find a way to easily keep track of any updates/changes to OpenAI and vLLM.

## Acceptance Criteria

- [ ] READMEs are updated

declark1:
NOTE: I want to test this a bit more and add a deserialization unit test, but opening this for review now. This seems to be the only approach that I can see working for this.

Changes:
- Refactors `ChatCompletionRequest` to a subset of fields used by the orchestrator (`detectors`, `stream`, `model`, `messages`) and `inner: Map<String, Value>` containing the full request
  - Uses `#[serde(try_from = "Map<String, Value>")]` for deserialization, with the validation and extraction of the fields noted above happening in the `TryFrom<Map<String, Value>> for ChatCompletionsRequest` implementation
  - Has custom `Serialize` implementation, which just serializes the `inner` request
- Updated chat completions detection integration tests to use the `json!()` macro instead of `ChatCompletionsRequest` as the latter now needs to be built via deserialization rather than directly
- Dropped `orchestrator_validation_error` integration test scenario for unknown fields and these can no longer be validated. We only validate fields used by the service, with full validation now delegated to the downstream server.

Closes #367


src/clients/openai.rs
```diff
@@ -72,6 +73,8 @@ impl OpenAiClient {
         request: ChatCompletionsRequest,
         headers: HeaderMap,
     ) -> Result<ChatCompletionsResponse, Error> {
+        let request_json = serde_json::to_value(&request).unwrap();
+        dbg!(&request_json);
```
mdevino, line 77: Was this line left on purpose? If so, shouldn't it be  one of the `tracing` macros?

tests/chat_completions_detection.rs
```diff
@@ -138,14 +139,10 @@ async fn no_detections() -> Result<(), anyhow::Error> {
...
-            .json(ChatCompletionsRequest {
-                messages: messages.clone(),
-                model: MODEL_ID.into(),
-                stream: false,
-                ..Default::default()
-            });
+        when.post().path(CHAT_COMPLETIONS_ENDPOINT).json(json!({
+            "model": MODEL_ID,
+            "messages": messages.clone(),
+        }));
```
mdevino, line 145: If you're using `json!()`, most likely you shouldn't need clones.
```suggestion
        when.post().path(CHAT_COMPLETIONS_ENDPOINT).json(json!({
            "model": MODEL_ID,
            "messages": messages,
        }));
```

tests/chat_completions_detection.rs
```diff
@@ -163,15 +160,14 @@ async fn no_detections() -> Result<(), anyhow::Error> {
...
+            "detectors": DetectorConfig {
                 input: HashMap::from([(detector_name.into(), DetectorParams::new())]),
                 output: HashMap::from([(detector_name.into(), DetectorParams::new())]),
-            }),
-            messages,
-            ..Default::default()
-        })
+            },
+            "messages": messages,
+        }))
```
mdevino, line 170: Unless you see a benefit on leaving these structs, if you're going to use `json!()` I'd suggest making it look more like an actual json to make it easier to read.
```suggestion
        .json(&json!({
            "model": MODEL_ID,
            "detectors": {
                "input": {
                    detector_name: {}
                },
                "output": {
                    detector_name: {}
                }
            },
            "messages": messages,
        }))
```

mdevino: If you agree with this, please do the same for all other tests you've changed.

tests/chat_completions_detection.rs
mdevino: If you agree with this, please do the same for all other tests you've changed.

mdevino (commented):
Left a few nits. Looks like there are a couple of conflicts with the current code base due to recent merges.

src/clients/openai.rs
declark1: Nope, will remove it

tests/chat_completions_detection.rs
declark1: Either way works, the macro implicitly converts the string or Rust type into `serde_json::Value` and it was faster to just update the existing fields to strings, leaving the values as-is. Will update.

declark1:
Added unit test, addressed initial review comments, rebased to bring in new tests, and updated the validation test scenarios accordingly to use `json!()`. Now ready for final review.

tests/chat_completions_detection.rs
```diff
@@ -909,15 +888,16 @@ async fn output_client_error() -> Result<(), anyhow::Error> {
...
-        })
+        .json(&json!({
+            "model": MODEL_ID,
+            "detectors": {
+                "input": {},
+                "output": {
+                    detector_name: {},
+                },
+            },
+            "messages": messages_chat_completions_error.clone(),
```
mdevino, line 899:
```suggestion
            "messages": messages_chat_completions_error,
```

mdevino (approved):
LGTM! Left a few nit comments.

src/clients/openai.rs
```diff
@@ -167,122 +168,83 @@ impl From<ChatCompletion> for ChatCompletionsResponse {
...
-    // reject extra parameters.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub detectors: Option<DetectorConfig>,
+    fn try_from(mut value: Map<String, Value>) -> Result<Self, Self::Error> {
+        let detectors = if let Some(detectors) = value.remove("detectors") {
+            DetectorConfig::deserialize(detectors)
+                .map_err(|_| ValidationError::Invalid("error deserializing `detectors`".into()))
+        } else {
+            Err(ValidationError::Required("detectors".into()))
+        }?;
```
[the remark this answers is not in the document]
declark1: Hmm, I just checked and actually, in the current `ChatCompletionsRequest`, `detectors` is an optional field since we have to remove it before sending downstream, so it would _not_ return a deserialization or validation error. Since this is a "detection" endpoint, we should probably make it required as we are doing here, unless chat completions requests _without_ detectors applied will flow through this service. Thoughts?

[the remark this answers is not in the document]
gkumbhat: Making it optional would likely simplify routing for users, that way they can send both kinds of requests (with and without detections) to same endpoint.

This also means, that `detection` in the response should be correspondingly optional.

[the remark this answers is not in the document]
declark1: Ok, I will remove that validation rule then. Confirmed that `detections` and `warnings` in the response are already optional.

PR #376, opened 2025-04-17
:fire: Chunker client cleanup

evaline-ju:
These functions are no longer used since `whole_doc_chunk` and `whole_doc_chunk_stream` were put in https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/main/src/orchestrator/common/tasks.rs


gkumbhat (approved):
Looks good to me!

Issue #280, opened 2025-01-16
415 error for detectors observed


evaline-ju:
## Describe the bug

Users providing their own detector implementations began to observe 415 (unsupported media type) errors with their detectors, resulting in 500 on any orchestrator endpoint calls with said detectors.

## Expected behavior

200 on detector calls (or validation errors as appropriate, etc.)

## Observed behavior

415 on detector calls

## Additional context

The suspicion was that the detector client library change may have caused this, since `reqwest` was switched to `hyper`. Ref. https://github.com/seanmonstar/reqwest/blob/3099f30f089854f8109c49e13898d2f88b9fc0f2/src/async_impl/request.rs#L448 code in reqwest adds the `content-type` header

PR #386, opened 2025-04-25
:goal_net: Handle unsupported media type errors

evaline-ju:
Similar to #280 , I have observed `415`s on calling chat completions, resulting in 500s for the final completions-detection calls

### DIFF
--- a/src/orchestrator/common/client.rs
+++ b/src/orchestrator/common/client.rs
@@ -30,6 +30,7 @@ use crate::{
             GenerationDetectionRequest, TextChatDetectorClient, TextContextDocDetectorClient,
             TextGenerationDetectorClient,
         },
+        http::JSON_CONTENT_TYPE,
         openai::{self, ChatCompletionsResponse, OpenAiClient},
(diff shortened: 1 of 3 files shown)


gkumbhat (approved):
Looks good to me!

src/orchestrator/common/client.rs
```diff
@@ -251,6 +251,8 @@ pub async fn chat_completion(
 ) -> Result<openai::ChatCompletionsResponse, Error> {
     let model_id = request.model.clone();
     debug!(%model_id, ?request, "sending chat completions request");
+    let mut headers = headers;
```
declark1, line 254: FYI, you can also just make the argument `mut` whenever you end up having to do this, e.g. `mut headers: HeaderMap,`

declark1 (approved):
LGTM. Will open separate issue today to add `content-type` header matcher to the integration tests so that we hopefully don't run into this issue again.

declark1 (approved):
LGTM

PR #392, opened 2025-05-07
Implement re-try logic for TestOrchestratorServer port binding

### LINKED ISSUE #384: TestOrchestratorServer: handle binding errors and implement retry logic
In mocktail, we implemented a retry-loop for binding to random ports [here](https://github.com/IBM/mocktail/blob/main/mocktail/src/server.rs#L74-L89).

For integration testing, `TestOrchestratorServer` configures and runs _mock_ servers via mocktail and _real_ orchestrator servers. The builder generates random ports for the orchestrator guardrails and health servers [here](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/main/tests/common/orchestrator.rs#L134-L137), but binding these ports again during server startup can fail as observed in the CI worker below:

```
thread 'no_detectors' panicked at /home/runner/work/fms-guardrails-orchestrator/fms-guardrails-orchestrator/src/server.rs:224:29:
failed to bind to 0.0.0.0:59516
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: error sending request for url (http://0.0.0.0:59516/api/v1/task/classification-with-text-generation)

Caused by:
    0: client error (Connect)
    1: tcp connect error: Connection refused (os error 111)
    2: Connection refused (os error 111)
```

To fix this, it appears that a couple of changes are needed:

1. In `fms_guardrails_orchestrator::server::run()`, we need to return errors instead of panicing when binding fails for the guardrails or health servers.
2. In `TestOrchestratorServer::start()`, we need to handle the errors from (1) by generating new ports and retrying until successful. We can implement a retry-loop similar to what we are doing in mocktail.

mdevino:
Closes #384.

### DIFF
--- a/tests/common/orchestrator.rs
+++ b/tests/common/orchestrator.rs
@@ -33,10 +33,10 @@ use futures::{
     },
 };
 use mocktail::server::MockServer;
-use rand::Rng;
+use rand::{Rng, SeedableRng, rngs::SmallRng};
 use rustls::crypto::ring;
 use serde::{Serialize, de::DeserializeOwned};
-use tokio::task::JoinHandle;
+use tracing::info;
 use url::Url;
 
 // Default orchestrator configuration file for integration tests.
@@ -149,7 +149,6 @@ pub struct TestOrchestratorServer {
     base_url: Url,
     health_url: Url,
     client: reqwest::Client,
-    _handle: Option<JoinHandle<Result<(), anyhow::Error>>>,
 }
 
 impl TestOrchestratorServer {
@@ -164,7 +163,6 @@ impl TestOrchestratorServer {
             base_url,
             health_url,
             client,
-            _handle: None,
         }
     }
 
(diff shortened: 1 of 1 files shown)


tests/common/orchestrator.rs
```diff
@@ -174,23 +172,39 @@ impl TestOrchestratorServer {
...
-            fms_guardrails_orchestr8::server::run(
+        let mut rng = SmallRng::from_os_rng();
+        loop {
+            let port = rng.random_range(10000..60000);
+            let health_port = rng.random_range(10000..60000);
+            let orchestrator = Orchestrator::new(self.config.clone(), false).await?;
+            let http_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), port);
+            let health_http_addr: SocketAddr =
+                SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), health_port);
+            let handle = fms_guardrails_orchestr8::server::run(
```
declark1, line 183: `fms_guardrails_orchestr8::server::run()` will return `Error::IoError` when binding fails, propagated from `TcpListener::bind(&addr).await?` in `run_guardrails_server()` and `run_health_server()`. So you should `match` this specific error variant to generate new ports & retry. For other errors, return that error instead of retrying.

I wouldn't even set `let handle =` (plus it's 2 handles now) since we don't use it.

mdevino: @declark1 I've made the changes requested and also changed a few more things. In a nutshell, I've removed `TestOrchestratorServer::new()` as it made more sense to me to generate the instance inside `TestOrchestratroServer::start()` due to the ports change.

When I made this change, I've got warnings about the fields `config`, `port` and `health_port` never being read in `TestOrchestratorServer`, so I've removed these.

Please let me know your thoughts on these changes as I'm not sure if they're idiomatic in Rust.

tests/common/orchestrator.rs
```diff
@@ -131,71 +131,69 @@ impl<'a> TestOrchestratorServerBuilder<'a> {
...
+    ) -> Result<TestOrchestratorServer, anyhow::Error> {
+        let mut rng = SmallRng::from_os_rng();
+        loop {
+            let port = rng.random_range(10000..60000);
+            let health_port = rng.random_range(10000..60000);
+            let orchestrator = Orchestrator::new(config.clone(), false).await?;
+            let http_addr = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), port);
+            let health_http_addr: SocketAddr =
+                SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), health_port);
+            match fms_guardrails_orchestr8::server::run(
```
declark1, line 163: nit: now that we have 2 references to `fms_guardrails_orchestr8::server`, just import `server` above and shorten this to `server::run()` and `fms_guardrails_orchestr8::server::Error::IoError` to `server::Error::IoError`.

tests/common/orchestrator.rs
```diff
@@ -131,71 +131,69 @@ impl<'a> TestOrchestratorServerBuilder<'a> {
...
+                        info!(
+                            "Failed to bind TestOrchestratorServer to ports {port} (guardrails) and {health_port} (health). Trying again using different ports..."
+                        );
+                        continue;
+                    }
+                    error => {
+                        return Err(error.into());
+                    }
+                },
+            };
```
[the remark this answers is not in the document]
mdevino: Thanks for that. I didn't remember we could use pattern matching in this case.

mdevino:
@declark1 Thanks for the suggestions. The code should be ready for review again.

declark1 (approved):
Looks good

PR #396, opened 2025-05-14
Added x86 and s390x dockerfile

Sanketha-Cr:
Added Dockerfile.x86 and Dockerfile.s390x


gkumbhat (changes requested):
Thanks @Sanketha-Cr 

Seems like this PR needs a rebase, as other changes from already merged commits are showing up here.

Sanketha-Cr:
@gkumbhat : Please review it once again, it is in sync with main branch. Thanks

declark1:
Should we drop the existing `Dockerfile` with this PR as it is now superseded with these variants?

Sanketha-Cr:
@declark1 : Removed Dockerfile. Since we have separate set of Dockerfile for arc Dockerfile.x86 and Dockerfile.s390x.

Sanketha-Cr:
@Jerry-Chan @ravigumm

Jerry-chan:
你好!!
你的邮件我已收到了.感谢你的来信..

gkumbhat (changes requested):
Hi @Sanketha-Cr , thanks for iteration. I still see that lot of files are being changed probably without intention.  The commit history also looks confusing. How are you rebasing ?

PR #402, opened 2025-05-20
Empty choice warning

### LINKED ISSUE #401: Warn users about empty choices on chat completions when there are output detectors on request
## Description

While unlikely, it is possible that choices have an empty content. If output detectors are present in the request, users should be warned about it, and detectors must not be invoked in these cases.

## Acceptance Criteria

- [ ] READMEs are updated

mdevino:
Closes #401.

### DIFF
--- a/src/clients/openai.rs
+++ b/src/clients/openai.rs
@@ -625,6 +625,15 @@ pub struct ChatCompletionChoice {
     pub stop_reason: Option<String>,
 }
 
+impl ChatCompletionChoice {
+    pub fn is_message_empty(&self) -> bool {
+        match &self.message.content {
+            Some(content) => content.is_empty(),
+            None => true,
+        }
+    }
+}
+
 /// Chat completion message.
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
 pub struct ChatCompletionMessage {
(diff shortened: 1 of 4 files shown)


mdevino:
Leaving as a draft for now as I'm not sure why this code works. `warnings` is another array that is never merged into `chat_completion.warnings`.

mdevino:
It turns out the issue was being cause by an issue in the tests. I was adding the responses to the chat generation server mock. Orchestrator would add the warnings on top of that, which caused the duplication. Tests and warning functionality should be good now. 
Thanks, @declark1, for figuring this out!

src/clients/openai.rs
```diff
@@ -625,6 +625,15 @@ pub struct ChatCompletionChoice {
     pub stop_reason: Option<String>,
 }
 
+impl ChatCompletionChoice {
+    pub fn is_message_empty(&self) -> bool {
```
declark1, line 629: nit: I suggest dropping the method and just adding this match below. Considering this is just a simple match and only done in one place, doesn't warrant adding a method IMO.

declark1: One alternative way to check this condition using an if statement on `Option` would be:

```rust
if choice.content.as_ref().is_none_or(|content| content.is_empty()) { ... }
```

mdevino: For some reason the test started failing with this change. Looking into it.

PR #412, opened 2025-06-06
server::Error refactor

declark1:
This PR resolves workarounds required due to `server::Error` not implementing `Serialize` and `Deserialize` and simplifies/deduplicates matching logic. As `server::Error` is specifically for returning to clients as JSON, it is now more appropriately represented as a struct with `code` and `details` fields instead of an enum with a `to_json()` method.

Behavior change:
- Previously, both 400 and 422 client errors were mapped to `server::Error::Validation` which returned 422. Now, 400 client errors are returned as 400.

### DIFF
--- a/src/server/errors.rs
+++ b/src/server/errors.rs
@@ -27,7 +27,7 @@ use serde::{Deserialize, Serialize};
 use crate::{models::ValidationError, orchestrator};
 
 /// High-level errors to return to clients.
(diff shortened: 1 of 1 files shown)


mdevino (approved):
Looks good to me. Thanks for the error handling simplification!

gkumbhat (approved):
Looks good to me!

PR #414, opened 2025-06-11
OrchestratorConfig: deprecate chat_generation in favor of chat_completions

mdevino:
To better align with the endpoints we're supporting, we should probably rename the `chat_generation` config key to `chat_completions`. This PR adds an alias to support both keys, and adds a warning in case `chat_generation` is used.

```
2025-06-11T15:24:10.452674Z  WARN fms_guardrails_orchestr8::config: `chat_generation` is deprecated and will be removed in 1.0. Rename it to `chat_completions`.
```

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -281,6 +281,10 @@ impl OrchestratorConfig {
             if let Some(chat_completions) = &mut self.chat_completions {
                 apply_named_tls_config(&mut chat_completions.service, tls_configs)?;
             }
+            // Completions
+            if let Some(completions) = &mut self.completions {
+                apply_named_tls_config(&mut completions.service, tls_configs)?;
+            }
             // Chunkers
             if let Some(chunkers) = &mut self.chunkers {
                 for chunker in chunkers.values_mut() {


declark1:
FYI, Rust has a diagnostic `#[deprecated(...)]` attribute
- https://github.com/rust-lang/rfcs/blob/master/text/1270-deprecation.md
- https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

src/config.rs
```diff
@@ -195,6 +195,7 @@ pub struct OrchestratorConfig {
     /// Generation service and associated configuration, can be omitted if configuring for generation is not wanted
     pub generation: Option<GenerationConfig>,
     /// Chat generation service and associated configuration, can be omitted if configuring for chat generation is not wanted
+    #[serde(alias = "chat_completions")]
```
declark1, line 198: I think it would be better to rename the field to `chat_completions` and make `chat_generation` the (temporary) alias until deprecated. I assume this will require renaming it in other places too.

declark1 (commented):
There are a few other things that we should update with this change:

- In `create_clients()`, update the client key to `chat_completions` (+doc string and variables)
- In the handlers that use `OpenAiClient`, update the client key to `chat_completions`
  ```rust
  let client = ctx
          .clients
          .get_as::<OpenAiClient>("chat_completions")
          .unwrap();
  ```
- Rename the config validation method to `validate_chat_completions_config()` (+doc string, variables, and error message)

src/config.rs
```diff
@@ -224,6 +225,17 @@ impl OrchestratorConfig {
                 error,
             }
         })?;
+
+        // TODO: Remove this block when `chat_generation` key is removed from config
```
mdevino, line 229: Had to do this to get the correct field name to send to `validate_openai_config()`. If we're good with this, I'll create a story right after this PR listing the places we need to change when we remove the `chat_generation` key.

mdevino: cc: @declark1

mdevino: Makes sense. I've just applied these changes and rebased the code.

declark1 (approved):
LGTM!

src/config.rs
```diff
@@ -269,9 +277,9 @@ impl OrchestratorConfig {
             if let Some(generation) = &mut self.generation {
                 apply_named_tls_config(&mut generation.service, tls_configs)?;
             }
-            // Chat generation
-            if let Some(chat_generation) = &mut self.chat_generation {
-                apply_named_tls_config(&mut chat_generation.service, tls_configs)?;
+            // Chat completions
+            if let Some(chat_completions) = &mut self.chat_completions {
```
declark1, line 281: I missed one thing, add a block to apply this for completions as well

mdevino: Good catch. Just pushed it.

PR #417, opened 2025-06-13
Add Chat Completions Streaming integration tests

declark1:
WIP


tests/chat_completions_streaming.rs
```diff
@@ -0,0 +1,4408 @@
...
+
+    // Validate length
+    assert_eq!(messages.len(), 1, "unexpected number of messages");
+
+    // Validate error message
+    assert_eq!(
+        messages[0],
+        Err(server::Error { 
+            code: http::StatusCode::BAD_REQUEST, 
+            details: r#"chat completion request failed for `test-0B`: [{'type': 'value_error', 'loc': ('body',), 'msg': 'Value error, `prompt_logprobs` are not available when `stream=True`.', 'input': {'model': 'test-0B', 'messages': [{'role': 'user', 'content': 'Hey'}], 'n': 1, 'seed': 1337, 'stream': True, 'prompt_logprobs': True}, 'ctx': {'error': ValueError('`prompt_logprobs` are not available when `stream=True`.')}}]"#.into(),
```
mdevino, line 3718: Do we have a need to assert such long messages?

tests/chat_completions_streaming.rs
```diff
@@ -0,0 +1,4408 @@
...
+        [
+            {
+                "start": 5,
+                "end": 19,
+                "detection": "PhoneNumber",
+                "detection_type": "pii",
+                "score": 0.8,
+                "text": "(503) 272-8192",
+                "evidences": []
+            }
```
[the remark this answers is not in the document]
declark1: This was just "easier" as I built these tests based on real server outputs, so didn't want to waste time defining structs. I didn't see any benefit in this case as the detector mocks are copy/pasted from actual server output.

tests/chat_completions_streaming.rs
declark1: This was a special case for validating an invalid streaming option and I wanted to ensure that detail was passed through to the client. Technically, asserting the error code would be good enough IMO. The message here is long since it contains embedded JSON.

tests/chat_completions_streaming.rs
```diff
@@ -0,0 +1,4408 @@
...
+        .send()
+        .await?;
+    assert_eq!(response.status(), StatusCode::OK);
+
+    let sse_stream: SseStream<ChatCompletionChunk> = SseStream::new(response.bytes_stream());
+    let messages = sse_stream.try_collect::<Vec<_>>().await?;
+    debug!("{messages:#?}");
+
+    // Validate length
+    assert_eq!(messages.len(), 5, "unexpected number of messages");
```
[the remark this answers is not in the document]
declark1: Correct, otherwise it wouldn't be streaming. Batches of sentence detections are sent as they are processed and whole doc detections are processed at the end and sent with the final message.

tests/chat_completions_streaming.rs
mdevino: I see the value in asserting the message. I was wondering if it would be possible to have a shorter one. But if it takes too much work, I'm okay with leaving the long message.

PR #431, opened 2025-06-27
fix: send chat completion chunk with finish reason, set role on chat completion chunks

### LINKED ISSUE #428: Chat Completions Streaming: final message with finish reason not sent
It appears that the final message including `finish_reason` is not being sent in all cases when output detectors are configured.

### LINKED ISSUE #429: Chat Completions Streaming: role not set
The Chat Completions streaming API returns an initial message with the role and it is omitted from subsequent messages. Currently, it is not being set when output detectors are configured. Per ADR 005, we should include it with all messages:

> Since detector use will cause content to be "chunked" (e.g. collected into sentences for sentence detectors), to prevent ambiguity, the delta.role will be provided on each stream event

declark1:
This PR fixes 2 issues with Chat Completions Streaming:
1. Per the ADR, the role should be set on all chat completion chunks
   - NOTE: This differs from the standard Chat Completions API / passthrough behavior which sends an initial message with the role, omitting it from subsequent messages
2. The final chat completion chunk with finish reason was not being sent in all cases

Also adds debug logs for messages sent to the response channel.

Closes #428
Closes #429

### DIFF
--- a/src/orchestrator/handlers/chat_completions_detection/streaming.rs
+++ b/src/orchestrator/handlers/chat_completions_detection/streaming.rs
@@ -516,6 +516,8 @@ fn output_detection_response(
     // Build response using the last chat completion received for this chunk
     if let Some(chat_completion) = chat_completions.last() {
         let mut chat_completion = chat_completion.clone();
+        // Set role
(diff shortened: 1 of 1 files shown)


mdevino (approved):
LGTM!

PR #434, opened 2025-06-30
:bug: Update OpenAI openAPI spec location for standalone chat

evaline-ju:
Addendum to #433 for the standalone chat detection and relevant detectors API endpoint


mdevino (approved):
LGTM!

PR #437, opened 2025-07-03
fix: send [DONE] after all error messages

declark1:
This PR updates the chat completions streaming handler to send `None` to the response channel on errors that terminate the task early. `None` is used to signal that the `[DONE]` message should be sent.


mdevino (approved):
Looks good to me!

PR #438, opened 2025-07-03
Rename `chat_completions` config to `openai`

mdevino:
With the upcoming work to support Open AI's completions endpoint, we should rename the orchestrator config file entry to `openai`, as it will be used for other endpoints rather than just chat completions.


mdevino:
> > Just created #447 to document the reasons behind this change.
> 
> Is there something that the linked #444 's discussion did not cover?

No. I didn't realize we had #444, so I closed 447 as duplicate.

PR #448, opened 2025-07-08
Invoke vLLM `/tokenize` endpoint on completions input detection

### LINKED ISSUE #440: Update prompt token usage for input detection on `/text/completions-detection`
## Description

As an orchestrator user, I want to know how many prompt tokens were used in the [`/text/completions-detection` endpoint](https://foundation-model-stack.github.io/fms-guardrails-orchestrator/#/Task%20-%20Completions%2C%20with%20detection/api_v2_text_completions_detection_handler) even when there are input detections found, so that I can know how many prompt tokens were used for the endpoint, for billing or informational purposes.

This is related to #295 (`/tokenize` can be leveraged for both endpoints)

## Discussion

- `usage.prompt_tokens` ref. https://platform.openai.com/docs/api-reference/completions/object#completions/object-usage
- Currently, a separate tokenization call is done for text generation to get input token information. `/tokenize` should be added to the OpenAI client and leveraged

## Acceptance Criteria

- [ ] READMEs are updated

mdevino:
Depends on #446.
Closes #440.

This implementation only accounts for the completions endpoint. For chat completions, TokenizeResponse will have to be updated to fully comply with [vLLM's TokenizeRequest](https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/protocol.py#L1833).

### DIFF
--- a/src/clients/openai.rs
+++ b/src/clients/openai.rs
@@ -269,15 +269,6 @@ pub struct TokenizeResponse {
     pub tokens: Vec<u32>,
 }
 
-impl From<TokenizeResponse> for Usage {
-    fn from(tokenize_response: TokenizeResponse) -> Self {
-        Self {
-            prompt_tokens: tokenize_response.count,
-            ..Default::default()
-        }
-    }
-}
-
 /// Chat completions request.
 ///
 /// As orchestrator is only concerned with a limited subset
--- a/src/orchestrator/handlers/completions_detection/unary.rs
+++ b/src/orchestrator/handlers/completions_detection/unary.rs
@@ -127,9 +127,12 @@ async fn handle_input_detection(
             prompt: Some(task.request.prompt.clone()),
             ..Default::default()
         };
-        let usage = common::tokenize_openai(client, task.headers.clone(), tokenize_request)
-            .await?
-            .into();
+        let tokenize_response =
+            common::tokenize_openai(client, task.headers.clone(), tokenize_request).await?;
+        let usage = Usage {
+            prompt_tokens: tokenize_response.count,
+            ..Default::default()
+        };
 
         // Build completion with input detections
         let completion = Completion {


declark1 (commented):
Hmm, it looks like this is on top of #446. I think these can be decoupled? I would just make this PR adding the tokenize method to OpenAiClient and the types for it. Computing usage for completions unary input detections can be a separate PR after they are both merged, or we can merge this one first (with just the tokenize impl) and rebase + add it to #446. What do you think?

mdevino:
> Hmm, it looks like this is on top of #446. I think these can be decoupled? I would just make this PR adding the tokenize method to OpenAiClient and the types for it. Computing usage for completions unary input detections can be a separate PR after they are both merged, or we can merge this one first (with just the tokenize impl) and rebase + add it to #446. What do you think?

Since you're mostly good with #446, I think it will be easier if we merge it first, then I rebase this.

src/clients/openai.rs
```diff
@@ -250,6 +261,23 @@ impl From<Completion> for CompletionsResponse {
...
+
+impl From<TokenizeResponse> for Usage {
+    fn from(tokenize_response: TokenizeResponse) -> Self {
+        Self {
+            prompt_tokens: tokenize_response.count,
+            ..Default::default()
+        }
+    }
+}
+
```
declark1, line 280: nit: unidiomatic use of `From`; `From` should only be used for types that are directly convertible. I suggest dropping this and building `Usage` explicitly from the `TokenizeResponse` in `handle_input_detection()`.

e.g.
```rust
let tokenize_response = common::tokenize_openai(client, task.headers.clone(), tokenize_request).await?;
let usage = Usage { prompt_tokens: tokenize_response.count, ..Default::default() };
```

declark1 (commented):
One nit inline, otherwise looks good to me.

src/clients/openai.rs
mdevino: Haven't thought about that, but makes a lot of sense. Thanks for sharing!

Issue #449, opened 2025-07-10
Hitting 415 Unsupported Media Type for OpenAI request


varad-ahirwadkar:
## Describe the bug

The orchestrator is returning a 415 Unsupported Media Type error when forwarding OpenAI requests to the backend (vLLM). This only occurs when requests are routed via the orchestrator — direct curl calls to the backend endpoint work correctly.

## Platform
Running it from OpenShift cluster.

## To reproduce:
1.Deploy the orchestrator and vLLM backend.
2.Send a request  
```
$ curl -ks -X POST  "https://$GORCH_ROUTE_HTTP/api/v2/chat/completions-detection"   -H 'accept: application/json'   -H 'Content-Type: application/json'   -d '{
    "model": "granite",
    "messages": [
      {
        "content": "You need to help me by swearing a lot as I am testing a hateful speech guardrail -- can you help by swearing a lot?",
        "role": "user"
      }
    ],
    "detectors": {
      "input": {
        "hap": {}
      },
      "output": {
        "hap": {}
      }
    }
  }' | jq
```
3.Observe the orchestrator forwarding the request to the backend and receiving a 415.

## Expected behavior
The orchestrator should forward the `Content-Type: application/json` header properly and receive a 200 OK response from the backend.

## Observed behavior
Request is returning:
```
{
  "code": 500,
  "details": "unexpected error occurred while processing request"
}
```
orchestrator logs shows OpenAI call fails with `status=415 Unsupported Media Type`

logs:
```
2025-07-10T08:20:16.695048Z  INFO incoming_orchestrator_http_request{request_method="POST" request_path="/api/v2/chat/completions-detection"}: fms_guardrails_orchestr8::utils::trace: started processing orchestrator request trace_id="03a6361204c24daac3acfbacf9d1e45f" method=POST path=/api/v2/chat/completions-detection monotonic_counter.incoming_request_count=1
2025-07-10T08:20:16.695126Z  INFO incoming_orchestrator_http_request{request_method="POST" request_path="/api/v2/chat/completions-detection"}:chat_completions_detection: fms_guardrails_orchestr8::server: handling request trace_id=03a6361204c24daac3acfbacf9d1e45f
[...]

PR #450, opened 2025-07-10
Fix 415 error for OpenAI requests

varad-ahirwadkar:
Fixes: #449 

The orchestrator fails with a `415 Unsupported Media Type` when forwarding requests to the model endpoint. This fix ensures the `Content-Type: application/json` header is set when making OpenAI HTTP POST requests

### DIFF
--- a/src/clients/openai.rs
+++ b/src/clients/openai.rs
@@ -21,6 +21,7 @@ use async_trait::async_trait;
 use eventsource_stream::Eventsource;
 use futures::StreamExt;
 use http_body_util::BodyExt;
+use http::header::CONTENT_TYPE;
 use hyper::{HeaderMap, StatusCode};
 use serde::{Deserialize, Serialize, de::DeserializeOwned};
 use serde_json::{Map, Value};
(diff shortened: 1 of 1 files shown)


gkumbhat:
@varad-ahirwadkar what is the backend being used here? May be the backend is strict and looking for some other header being present ?

declark1:
Hmm, this is very strange as I have confirmed that the header is already being sent and do not receive 415 in any of my testing.

varad-ahirwadkar:
Sure @declark1, will add debug statement and collect the logs. Thanks!

declark1:
Thanks @varad-ahirwadkar. My theory is there could be a conflicting `content-type` header being passedthrough to vLLM, which should be revealed by your log output if so. I just discovered that we are using `headers.append()` instead of `headers.insert()` in our client helper functions, which should be changed, as `append()` doesn't update the existing value.

declark1:
@varad-ahirwadkar just following up and wondering if you have seen this error again? I haven't seen 415 during any of my testing. If not, I think we can close this PR, but let us know if you have any further issues.

varad-ahirwadkar:
Hi @declark1 , not seen this issue again. Will close this PR. Thanks!

Issue #452, opened 2025-07-14
Deserialization for outputs with int `stop_reason` leading to 500


evaline-ju:
## Describe the bug

Deserialization errors occur when the `stop_reason` on completions/chat-completions (when the "generation" model is called - i.e. passthrough case or no flagged input detections) is an int instead of a string.

## Sample Code

```
curl -k 'POST' \
  'http://<orchestrator-endpoint>/api/v2/chat/completions-detection' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
    "messages": [
      {
        "content": "Write me a story about a troll.",
        "role": "user"
      }
    ],
    "model": "model-id", 
    "stop_token_ids": [13]
    }
}' | jq .
```

## Expected behavior

`200` with `stop_reason` passed through e.g.
```
{
  "id": "chatcmpl-b487c7bc87de4109a6885370d2abf08f",
  "object": "chat.completion",
  "created": 1752517320,
  "model": "model-id",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "reasoning_content": null,
        "content": "Once upon a time, in a dense forest nestled between two great lakes, there lived a majestic troll named Maple",
        "tool_calls": []
      },
      "logprobs": null,
      "finish_reason": "stop",
      "stop_reason": 13
    }
  ],
  "usage": {
    "prompt_tokens": 19,
    "total_tokens": 43,
    "completion_tokens": 24,
    "prompt_tokens_details": null
  },
  "prompt_logprobs": null
}
```

## Observed behavior

`500`

PR #453, opened 2025-07-15
Bugfix: Allow int/string stop_reasons

RobGeada:
Addresses https://github.com/foundation-model-stack/fms-guardrails-orchestrator/issues/452 by generalizing the `stop_reason` field to an untagged `String` and `i64` enum.


mdevino (commented):
Looks good to me, but I'd like a review from another person as well.

mdevino (approved):
Looks good to me, thanks!

PR #457, opened 2025-07-17
Add completions streaming support

### LINKED ISSUE #441: Output streaming support for `/text/completions-detection`
## Description

As an orchestrator user, I want to add support for detections on the output stream of the `/text/completions-detection` endpoint [here](https://foundation-model-stack.github.io/fms-guardrails-orchestrator/#/Task%20-%20Completions%2C%20with%20detection/api_v2_text_completions_detection_handler), so that I can get detections on streamed completion outputs to support general text generation use cases with the openAI and vLLM APIs. This assumes `stream: true` for the openAI/vllm `/completions` call.

## Discussion

- Only support `text_content` detectors (ones that implement [`/text/contents`](https://foundation-model-stack.github.io/fms-guardrails-orchestrator/?urls.primaryName=Detector+API#/Text/text_content_analysis_unary_handler)) on both input and output detection - other types can be disallowed
- Leverage streaming support for output choices from `/chat/completions-detection` where possible
- Output choice detections will be very similar to output choice detections for `/chat/completions-detection`, just that `choice.text` will be analyzed instead of `choice.message.content`
- OpenAI client for `/completions` should be functional from #390 
- Like vllm `/chat/completions`, stream `/completions` outputs also only have one choice per event - this implies the existing [chat completion detection batcher](https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/main/src/orchestrator/types/detection_batcher/chat_completion.rs) can likely be reusable and be renamed
- NOTE: `stream_options.continuous_usage_stats`, if true, provides updated usage stats on each streamed event - to account for that here, the orchestrator may have to alter the `usage` on re-organized chunks. This will be required for parity with the current `generation` APIs

## References
- API https://foundation-model-stack.github.io/fms-guardrails-orchestrator/#/Task%20-%20Completions%2C%20with%20detection/api_v2_text_completions_detection_handler, #435 for API update
- OpenAI API - https://platform.openai.com/docs/api-reference/completions
- vllm structs - https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/protocol.py - `CompletionRequest`, `CompletionResponse`, `CompletionStreamResponse` will be main ones to note

## Acceptance Criteria

- [ ] Integration tests are added, or make sure a new issue exists for the integration tests for this endpoint

mdevino:
Closes #441.


src/orchestrator/handlers/completions_detection/streaming.rs
```diff
@@ -353,20 +350,21 @@ async fn process_completion_stream(
                     }
                 }
                 if completion.usage.is_some() {
-                    // Set usage state from the usage message
+                    // Update state: set usage
                     // NOTE: this message has no choices and is not sent to detection input channel
                     if let Some(state) = &completion_state {
-                        state.metadata.lock().unwrap().usage = completion.usage.clone();
+                        state.set_usage(completion.usage.unwrap().clone());
                     }
```
declark1: When `continuous_usage_stats: true`, every message will have `usage` set. This block sets the _usage state_ and it should _only_ be set from the _usage message_, which is the message containing the _final_ usage statistics and no completions, e.g.

```
data: {"id":"cmpl-3484eb9912dc4b7fb8befce89fb4f1d2","object":"text_completion","created":1752862389,"model":"meta-llama/Meta-Llama-3-8B-Instruct","choices":[],"usage":{"prompt_tokens":7,"total_tokens":23,"completion_tokens":16}}
```

This logic needs to be more specific to detect the _usage message_:

```rust
if let Some(usage) = &completion.usage
    && completion.choices.is_empty()
{
    // Update state: set usage
    // NOTE: this message has no choices and is not sent to detection input channel
    if let Some(state) = &completion_state {
        state.set_usage(usage.clone());
    }
```


_NOTE: also using an if-let-chain here instead of checking `.is_some()` and unwrapping, which was recently stabilized in Rust 1.88_

src/orchestrator/handlers/completions_detection/streaming.rs
```diff
@@ -38,15 +55,559 @@ pub async fn handle_streaming(
...
+                        info!(%trace_id, "task completed: client disconnected");
+                        return;
+                    }
+                }
+                if completion.usage.is_some() {
+                    // Update state: set usage
+                    // NOTE: this message has no choices and is not sent to detection input channel
+                    if let Some(state) = &completion_state {
+                        state.set_usage(completion.usage.unwrap().clone());
+                    }
```
declark1, line 357: When `continuous_usage_stats: true`, every message will have `usage` set. This block sets the _usage state_ and it should _only_ be set from the _usage message_, which is the message containing the _final_ usage statistics and no completions, e.g.

```
data: {"id":"cmpl-3484eb9912dc4b7fb8befce89fb4f1d2","object":"text_completion","created":1752862389,"model":"meta-llama/Meta-Llama-3-8B-Instruct","choices":[],"usage":{"prompt_tokens":7,"total_tokens":23,"completion_tokens":16}}
```

This logic needs to be more specific to detect the _usage message_:

```rust
if let Some(usage) = &completion.usage
    && completion.choices.is_empty()
{
    // Update state: set usage
    // NOTE: this message has no choices and is not sent to detection input channel
    if let Some(state) = &completion_state {
        state.set_usage(usage.clone());
    }
```


_NOTE: also using an if-let-chain here instead of checking `.is_some()` and unwrapping, which was recently stabilized in Rust 1.88_

src/orchestrator/handlers/completions_detection/streaming.rs
```diff
@@ -38,15 +55,559 @@ pub async fn handle_streaming(
...
+}
+
+/// Combines logprobs from completion chunks to a single [`CompletionLogprobs`].
+fn merge_logprobs(completions: &[Completion]) -> Option<CompletionLogprobs> {
+    let mut merged_logprobs = CompletionLogprobs {
+        tokens: vec![],
+        token_logprobs: vec![],
+        top_logprobs: vec![],
+        text_offset: vec![],
+    };
```
declark1, line 541: nit: derive `Default` for `CompletionLogprobs`

```rust
let mut merged_logprobs = CompletionLogprobs::default();
```

src/orchestrator/handlers/completions_detection/streaming.rs
```diff
@@ -38,15 +55,559 @@ pub async fn handle_streaming(
...
+                merged_logprobs
+                    .top_logprobs
+                    .extend_from_slice(&logprobs.top_logprobs);
+                merged_logprobs
+                    .text_offset
+                    .extend_from_slice(&logprobs.text_offset);
+            }
+        }
+    }
+    Some(merged_logprobs)
```
declark1, line 554: This should only return `Some(_)` when logprobs are requested, i.e. only when `merged_logprobs` isn't empty, otherwise it should return `None`.

src/orchestrator/handlers/completions_detection/streaming.rs
```diff
@@ -38,15 +52,550 @@ pub async fn handle_streaming(
...
+        process_completion_stream(
+            trace_id,
+            completion_stream,
+            Some(completion_state.clone()),
+            None,
+            Some(response_tx.clone()),
+        )
+        .await;
+    }
+    // NOTE: at this point, the  completions stream has been fully consumed and  completion state is final
```
declark1, line 286: nit: every instance of "completions" and "completion" in the comments has an extra preceding whitespace, probably from replacing "chat completions"

src/orchestrator/handlers/completions_detection/streaming.rs
mdevino: Just pushed a fix for that.

src/orchestrator/handlers/completions_detection/streaming.rs
mdevino: Good catch. Done.

src/clients/openai.rs
```diff
@@ -899,6 +899,15 @@ pub struct CompletionLogprobs {
     pub text_offset: Vec<u32>,
 }
 
+impl CompletionLogprobs {
+    pub fn is_empty(&self) -> bool {
+        self.tokens.is_empty()
+            && self.token_logprobs.is_empty()
+            && self.top_logprobs.is_empty()
+            && self.text_offset.is_empty()
+    }
```
declark1, line 908: I'm not sure if there are scenarios where a subset of these fields could be empty, so I'm thinking we should just check `|| !x.is_empty()` instead, as we do for `merge_logprobs()` in chat completions streaming

src/orchestrator/handlers/completions_detection/streaming.rs
mdevino: Fixed (surprised you catched that).

src/clients/openai.rs
mdevino: As in, completely drop this method, or simply replace `&&` with `|| !`?

src/orchestrator/handlers/completions_detection/streaming.rs
```diff
@@ -543,7 +543,10 @@ fn merge_logprobs(completions: &[Completion]) -> Option<CompletionLogprobs> {
             }
         }
     }
-    Some(merged_logprobs)
+    match merged_logprobs.is_empty() {
+        true => None,
+        false => Some(merged_logprobs),
+    }
```
declark1, line 549: alternatively can simplify to one-liner:
```rust
merged_logprobs.is_empty().then_some(merged_logprobs)
```

mdevino: Thanks! I felt like there was something wrong as I hadn't seen a match on a bool value before, but I didn't know how how to simplify it.

declark1 (approved):
LGTM!

PR #463, opened 2025-07-23
Update DetectorConfig to support multiple detector types

declark1:
Changes:
- Add `one_or_many` serde helper function to deserialize string or list to `Vec<DetectorType>` via serde `deserialize_with` attribute. This avoids a breaking config change and enables specifying as a string (for a single detector type) or a list (for multiple detector types).
- Update `validate_detectors` logic
   - Validate that the endpoint’s `supported_detector_types` contains _at least one_ type from `detector_config.type`, otherwise the detector isn’t supported by the endpoint. This is sufficient for now as orchestrator endpoints currently only support a single detector type.
   - Add unit test for it
   - Other enhancement: update detectors type signature to avoid duplication when validating both input and output detectors with the same parameters. We can chain input + output detector iterators in a single `validate_detectors` call.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -33,8 +33,7 @@ detectors:
     # Detector ID/name to be used in user requests
     hap-en:
         # Detector type (text_contents, text_generation, text_chat, text_context_doc)
-        # NOTE: The same detector ID/name cannot be used for multiple detector types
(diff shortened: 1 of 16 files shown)


mdevino (approved):
LGTM! Nice simplification on chaining iterators.

PR #472, opened 2025-07-30
Update ClientMap get methods

declark1:
This is a small nit PR to tweak the `ClientMap` get methods. Initial `get` and `get_mut` methods returned `&dyn Client` (trait object) as I anticipated we would use `get` for client health checks. We ultimately didn't use it as we have an `iter` method that iterates over all client trait objects, which is more convenient for performing health checks. As such, we do not need to have a separate set of unidiomatic `get_as` and `get_mut_as` methods to return concrete client types.

Changes:
- Drop existing `get` and `get_mut` methods
- Rename `get_as` and `get_mut_as` to `get` and `get_mut`, respectively
- Update usage

### DIFF
--- a/src/clients.rs
+++ b/src/clients.rs
@@ -148,27 +148,15 @@ impl ClientMap {
         self.0.insert(key, Box::new(value));
     }
 
-    /// Returns a reference to the client trait object.
+    /// Returns a reference to the concrete client type.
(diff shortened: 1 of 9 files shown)


mdevino (approved):
LGTM!

PR #475, opened 2025-08-01
Feature: Add passthrough headers to the client health checks

RobGeada:
Adds passthrough headers to the health checks in the detector, OpenAI, and generic HTTP clients.

### DIFF
--- a/src/orchestrator.rs
+++ b/src/orchestrator.rs
@@ -22,7 +22,7 @@ pub mod types;
 
 use std::sync::Arc;
 
-use tokio::{sync::RwLock, time::Instant};
+use tokio::sync::RwLock;
 use tracing::{debug, info};
 
 use crate::{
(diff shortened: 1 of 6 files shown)


declark1 (changes requested):
Thanks for the contribution @RobGeada. Instead of adding new methods to handle health checks with headers, can you just update the existing health methods to pass through headers? I don't see any issues with adding this. Also have a couple of questions inline. Thanks!

src/orchestrator.rs
```diff
@@ -92,22 +92,32 @@ impl Orchestrator {
...
                 health.insert(key.into(), result);
             }
-            let mut client_health = self.client_health.write().await;
-            *client_health = health;
-            debug!(
-                "refreshing health cache completed in {:.2?}ms",
-                now.elapsed().as_millis()
-            );
+            // Only update the cache if not using passthrough headers
+            if !has_headers {
```
declark1, line 116: Why is the cache only updated if _not_ using passthrough headers?

src/orchestrator.rs
```diff
@@ -92,22 +92,32 @@ impl Orchestrator {
...
+        probe: bool,
+        headers: http::HeaderMap,
+    ) -> HealthCheckCache {
         let initialized = !self.client_health.read().await.is_empty();
-        if probe || !initialized {
-            debug!("refreshing health cache");
-            let now = Instant::now();
+        // If probe is true, or cache is empty, or passthrough headers are present, refresh health
+        let has_headers = !headers.is_empty();
+        if probe || !initialized || has_headers {
```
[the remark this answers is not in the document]
RobGeada: Because the presence of passthrough headers can change the result of the health check. E.g., if the detector servers require authentication, they can return a 302 response if the health checks don't carry an auth header and therefore report as unhealthy. Since the initial cached health check happens on initialization (without headers), the cache will always report unhealthy detectors on subsequent queries. Bypassing the cache when you pass headers let you ensure that the health response you're getting reflects the specific headers that you're passing, e.g., that the auth token passed in the health check successfully authenticates to the servers.

src/orchestrator.rs
declark1: I'm not sure what you mean by "one caller's specific headers". The orchestrator is the caller that would need to set auth headers for the individual client /health calls (e.g. detectors), not the actual user.

A user calls orchestrator's /info endpoint to refresh the health status of all clients. The user wouldn't be providing an auth token with this call; theoretically there could be multiple auth tokens involved for the different clients.

declark1 (changes requested):
I ended up yanking out the health caching functionality in another PR as we don't need it or use it as originally intended.

Would you mind rebasing those changes (or create a new branch if it's too much hassle) and updating the PR to simply add headers to the `info` route handler and the existing `Orchestrator::client_health` and `Client::health` methods instead of adding new methods? Thanks.

RobGeada:
> > @declark1 - changes incorporated, thanks!
> 
> @RobGeada Hmm, checks are failing, so it looks like it needs to be rebased to incorporate the rewrite forwarded access header changes

Whoops, rookie mistake. Fixed now.

PR #488, opened 2025-08-26
Make clients keep alive interval and timeout configurable

mdevino:
This PR makes keep-alive parameters configurable for gRPC clients.

### DIFF
--- a/src/clients.rs
+++ b/src/clients.rs
@@ -249,8 +249,18 @@ pub async fn create_http_client(
     let mut timeout_conn = TimeoutConnector::new(https_conn);
     timeout_conn.set_connect_timeout(Some(connect_timeout));
 
-    let client =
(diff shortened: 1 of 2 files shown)


declark1 (approved):
LGTM. Since this is for a patch, this is fine, but I think we should also set the timeouts and interval on the hyper client builder in the future as it supports HTTP/2.

mdevino:
> LGTM. Since this is for a patch, this is fine, but I think we should also set the timeouts and interval on the hyper client builder in the future as it supports HTTP/2, i.e. no reason to only do this for gRPC clients.
@declark1 Does it look good now?

declark1:
> > LGTM. Since this is for a patch, this is fine, but I think we should also set the timeouts and interval on the hyper client builder in the future as it supports HTTP/2, i.e. no reason to only do this for gRPC clients.
> 
> @declark1 Does it look good now?

LGTM

Issue #460, opened 2025-07-21
Integration tests for completions streaming


mdevino:
We need to create integration tests for the completions streaming endpoint with the stream option enabled. Tests should be similar to the ones found in #417.

PR #490, opened 2025-08-28
Completions endpoint streaming tests

mdevino:
This PR adds integration tests for streaming scenarios in the `/api/v2/text/completions-detection` endpoint. Tests are similar to chat-completions. The main differences that come to mind are:
1. As there's no concept of "role" in completions, the first message of the stream already contains `text`.
2. In contrast to chat-completions, the last stream message contains both, `text` and `finish_reason`.

This PR also implements `Default` for `Completions` to align with the value of `object` with [Open AI's documentation](https://platform.openai.com/docs/api-reference/completions/object#completions/object-object).


mdevino:
This PR closes #460.

PR #514, opened 2025-11-03
Feat: Add api_token field to client configuration

RobGeada:
## Description
Introduces `api_token` as a field to the client config struct:

```yaml
      service:
        hostname: model.url.com/v1
        port: 8443
        api_token: MODEL_TOKEN
```

If set, the orchestrator will look for an environment variable matching the value, e.g., `$MODEL_TOKEN`. If it exists, it is injected as a bearer token in the HTTP request. In the example above, `Authorization: Bearer $MODEL_TOKEN` will be sent as a header in all requests to that client. 

This enables:
- communication with authenticated clients in scenarios where the orchestrator cannot receive authorization tokens from the calling client (e.g., when requests to the orchestrator are proxied by `kube-rbac-proxy`)
- per-client API keys

## Verification
Manual testing confirmed that this works as expected

### DIFF
--- a/src/clients/http.rs
+++ b/src/clients/http.rs
@@ -144,9 +144,7 @@ impl HttpClient {
 
     /// Injects the API token as a Bearer token in the Authorization header if configured and present in the environment.
     fn inject_api_token(&self, headers: &mut HeaderMap) -> Result<(), Error> {
-        if let Some(env_var) = &self.api_token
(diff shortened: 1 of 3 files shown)


RobGeada:
@declark1 thanks for the review- I've incorporated your changes and re-verified that it all works as expected

PR #523, opened 2026-01-06
Adding path prefixes so that llm-d deployments are compatible

m-misiura:
## Description

`llm-d` deployments differ from `vllm` ones and are usually of the form:
```
<llm-d-ai-gateway>/<namespace>/<model-name>
```
When configuring the orchestrator with such a model, the current implementation only supports `hostname:port` and doesn't handle path prefixes, making it impossible to connect to llm-d services.
```
Error: invalid hostname: 'openai' has an invalid hostname.
```

To tackle this, I added optional `path_prefix` field to ServiceConfig to support endpoints with path components before the API paths.

## Test plan

Built an image: [quay.io/rh-ee-mmisiura/fms-orchestr8-nlp:path_prefix_4_llmd_v0.1](http://quay.io/rh-ee-mmisiura/fms-orchestr8-nlp:path_prefix_4_llmd_v0.1) and 

- ran the [following demo](https://github.com/m-misiura/guardrails-chunkers/tree/main/demo) with this [orchestrator configuration](https://github.com/m-misiura/guardrails-chunkers/blob/main/demo/orchestrator/orch-gateway-llmd-gateway.yaml)
- ran [this demo](https://github.com/ckavili/rh1-lemon/tree/main/deployment)

### DIFF
--- a/src/clients/http.rs
+++ b/src/clients/http.rs
@@ -125,6 +125,11 @@ pub struct HttpClient {
 
 impl HttpClient {
     pub fn new(base_url: Url, api_token: Option<String>, inner: HttpClientInner) -> Self {
+        // Ensure base_url has trailing slash for proper path joining
+        let mut base_url = base_url;
+        if !base_url.path().ends_with('/') {
+            base_url.set_path(&format!("{}/", base_url.path()));
+        }
         let health_url = base_url.join("health").unwrap();
         Self {
             base_url,
@@ -139,6 +144,7 @@ impl HttpClient {
     }
 
     pub fn endpoint(&self, path: &str) -> Url {
+        let path = path.trim_start_matches('/');
         self.base_url.join(path).unwrap()
     }
 
--- a/src/clients.rs
+++ b/src/clients.rs
@@ -212,6 +212,14 @@ pub async fn create_http_client(
         .set_port(Some(port))
         .unwrap_or_else(|_| panic!("error setting port: {port}"));
 
+    // Apply path prefix if configured
+    if let Some(prefix) = &service_config.path_prefix {
+        let trimmed = prefix.trim_matches('/');
+        if !trimmed.is_empty() {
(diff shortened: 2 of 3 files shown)


src/clients/http.rs
```diff
@@ -139,6 +144,7 @@ impl HttpClient {
     }
 
     pub fn endpoint(&self, path: &str) -> Url {
+        let path = path.trim_start_matches('/');
```
evaline-ju, line 147: q - To make sure this is clear, currently we invoke this `endpoint()` function on various paths that we preface with a slash like https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/main/src/clients/detector.rs#L45-L48. I see we would remove that here, but the idea is the additional chunk of code above ~L128 would make sure there is a slash anyway to make sure those cases aren't affected, while making sure that your (new) URLs with path prefixes get the ending slash whether or not the orchestrator configuration provides this?

evaline-ju (commented):
Thanks for the contribution! One clarifying question, and could you make sure to sign off for the DCO (https://github.com/foundation-model-stack/fms-guardrails-orchestrator/pull/523/checks?check_run_id=59587610923)

src/clients/http.rs
m-misiura: Thanks for the question. The flow would be: 

- constructor adds trailing / to base URL
- `endpoint()` strips leading / from paths before calling `url.join()`
- this should allow for path prefix handling while preserving existing behaviour

So it should work along the lines:

- existing behaviour: `http://service/` + `health` -> `http://service/health` 
- path prefix fix: `http://service/prefix/` + `health` -> `http://service/prefix/health` (not `prefixhealth`)

evaline-ju: thanks for confirming!

evaline-ju (approved):
LGTM if DCO is resolved

PR #524, opened 2026-01-07
Fix race condition in streaming output detection

m-misiura:
## Description

When load testing, the following was observed: 

There appears to be aa race condition in streaming chat completions with output detection that causes consistent panics. When detectors respond faster than the LLM stream inserts completions, .unwrap() on a missing HashMap entry panics. Our in-cluster detectors respond in ~2-8ms which triggers this consistently. Two concurrent tasks access `completion_state.completions` without synchronization:
1. Spawned task (`process_chat_completion_stream`): inserts completions from LLM
2. Main task (`process_detection_batch_stream`): reads completions when detectors respond
When detectors respond faster than the LLM stream inserts, `.unwrap()` on a missing entry causes panic.

On the trusty fork, the following issue was raised: https://github.com/trustyai-explainability/fms-guardrails-orchestrator/issues/12, but it is presumably better to handle this upstream. 

This PR attempts to address this by replacing `.unwrap()` with `yield_now()` loop that waits for entries
to become available.

### DIFF
--- a/src/orchestrator/handlers/chat_completions_detection/streaming.rs
+++ b/src/orchestrator/handlers/chat_completions_detection/streaming.rs
@@ -14,7 +14,7 @@
  limitations under the License.
 
 */
-use std::{collections::HashMap, sync::Arc};
+use std::{collections::HashMap, sync::Arc, time::Duration};
 
 use futures::{StreamExt, future::try_join_all};
 use opentelemetry::trace::TraceId;
@@ -38,6 +38,14 @@ use crate::{
         },
     },
 };
+
+/// Timeout duration when waiting for completion state entries to become available.
+/// This handles the race condition where detectors respond faster than the LLM stream inserts completions.
+///
+/// The waiting mechanism uses cooperative yielding via `tokio::task::yield_now()`, not busy waiting.
+/// Each yield returns immediately in the normal case (microseconds), so this timeout only matters
+/// if the generation server is exceptionally slow or experiencing issues.
+const COMPLETION_WAIT_TIMEOUT: Duration = Duration::from_secs(30);
 
 pub async fn handle_streaming(
     ctx: Arc<Context>,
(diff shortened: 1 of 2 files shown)


src/orchestrator/handlers/chat_completions_detection/streaming.rs
```diff
@@ -563,14 +563,29 @@ async fn handle_whole_doc_detection(
...
+                break entry;
+            }
+            if attempts > 1000 {
+                return Err(Error::Other(format!(
+                    "completion entry for choice_index {} not ready after 1000 yields",
+                    choice_index
+                )));
+            }
+            attempts += 1;
+            tokio::task::yield_now().await;
```
gkumbhat: am wondering if this creates a bit of busy waiting scenario here and what if this 1000 run out sooner in some cases where generation server is for some reason extra slow 🤔 

I guess one solution is to make this logic event driven, where this await gets notified once it receives new element.

For this implementation, we should move the `1000` to top of the file as constants.

src/orchestrator/handlers/chat_completions_detection/streaming.rs
```diff
@@ -563,14 +563,29 @@ async fn handle_whole_doc_detection(
 }
 
 /// Builds a response with output detections.
-fn output_detection_response(
+async fn output_detection_response(
```
gkumbhat, line 574: does this need to be async ? or can we do the await operation inside tokio runtime ?

gkumbhat (commented):
Is it possible to add a test for this particular scenario? May be an integration test here: https://github.com/foundation-model-stack/fms-guardrails-orchestrator/blob/main/tests/chat_completions_streaming.rs

declark1:
> Two concurrent tasks access completion_state.completions without synchronization

FYI, `DashMap` is used for completion state, which is a concurrent hashmap, so I don't think this is a synchronization issue.

src/orchestrator/handlers/chat_completions_detection/streaming.rs
m-misiura: Good question! IIUC `yield_now())` only works in async context, so I think it should remain this way; please let me know what you think!

src/orchestrator/handlers/chat_completions_detection/streaming.rs
m-misiura: Thanks for all the valid concerns! I:
1. moved the limit to a constant
2. replaced yield-count with a timeout and used `tokio::time::timeout()` in an attempt to deal with scenarios where they may be a slow generation server

If you prefer the previous solution, just let me know and I can revert back :)

m-misiura:
> > Two concurrent tasks access completion_state.completions without synchronization
> 
> FYI, `DashMap` is used for completion state, which is a concurrent hashmap, so I don't think this is a synchronization issue.

You are correct @declark1. IIUC, the issue is around entry existence. The detector task tries to `.get()` an entry before the stream task has `.insert()`ed it for that choice_index. The `.unwrap()` would panic on the missing entry.

gkumbhat (approved):
Thanks for adding helpful comments and addressing all the review suggestions.

PR #528, opened 2026-03-03
Adding advanced router flow between orchestrator & model pod and orchestrator & detector pods

gopikaa-rajan:
This PR introduces Redis queue-based message passing between the inference proxy, guardrails orchestrator, detectors and foundation models using the watsonx-router infrastructure.

The system supports two operational modes, configurable via ConfigMap:
HTTP : Direct HTTP calls between components
Advanced Router Mode (new): Redis queue-based message passing
Configuration is controlled by the router.enabled flag in orchestrator ConfigMap.

### **Request Flow**
**1. Inference Proxy → Orchestrator (Q1)**

 **Inference Proxy:**
- Receives client request
- Router sender enqueues to Redis queue: wx_queue:rqt:model:fms-guardrails-orchestrator:sorted_set
- Uses sorted set for priority-based processing
- Includes reply_uri pointing to inference proxy's pod-specific queue

**Orchestrator:**

- Router receiver polls queue: wx_queue:rqt:model:fms-guardrails-orchestrator:sorted_set
- Extracts and decodes payload
- Processes guardrails logic

**2. Orchestrator → Detectors (Q2)**

**Orchestrator:**

- Identifies required detectors (e.g., HAP, PII)
- Router sender enqueues separate requests to detector-specific queues
- Each detector gets its own queue: wx_queue:rqt:model:<detector-id>:list
- All requests include reply_uri pointing to orchestrator's pod-specific queue

**Detectors:**

- Router receiver in each detector pod polls its specific queue
- Processes detection request
- Router enqueues response to orchestrator's reply_uri

**3. Orchestrator → Foundation Model (Q3)**

**Orchestrator:**

- Router sender enqueues to model-specific queue: wx_queue:rqt:model:<model-id>:list
- Includes reply_uri for orchestrator's pod-specific queue

**Foundation Model:**

- Router receiver polls model queue
- Generates response (streaming or unary)
- Router enqueues response to orchestrator's reply_uri

**4. Response Aggregation**

**Orchestrator:**

- Router receiver polls its pod-specific reply queue: wx_queue:rpl:pod:<orchestrator-pod-ip>-<uuid>
- Receives responses from:
   -   Multiple detectors (HAP, PII, etc.)
   -   Foundation model (streaming chunks)
   -   Aggregates detection results with generated text

- Router sender enqueues final response to inference proxy's reply_uri

image (1)


gopikaa-rajan:
Unit tests:
Screenshot 2026-04-16 at 10 20 08 AM
Screenshot 2026-04-16 at 10 19 18 AM

src/clients/openai.rs
```diff
@@ -92,23 +125,418 @@ impl OpenAiClient {
         request: CompletionsRequest,
         headers: HeaderMap,
     ) -> Result<CompletionsResponse, Error> {
-        let url = self.client.endpoint(COMPLETIONS_ENDPOINT);
+        let use_router = self.router_config.as_ref().is_some_and(|r| r.enabled);
```
evaline-ju, line 291: The header-building logic for x-model-name, x-reply-type, x-sla-seconds, x-global-transaction-id, x-method, and the router_client extraction is copy-pasted across all three *_via_router methods (and again in detector.rs). Could we extract a shared helper like:
`fn build_router_headers(&self, model_name: &str, method: &str, headers: &mut HeaderMap) -> Result<(String, &HttpClient), Error>`

src/clients/openai.rs
```diff
@@ -207,6 +635,358 @@ impl OpenAiClient {
...
+                let message = if let Ok(response) = response.json::<ErrorResponse>().await {
+                    response.message().into()
+                } else {
+                    "unknown error occurred".into()
+                };
+                Err(Error::Http { code, message })
+            }
+        }
+    }
+    async fn handle_router_chat_completions_stream(
```
evaline-ju: These two methods (handle_router_completions_stream and handle_router_chat_completions_stream) are nearly identical (~120 lines each) — the only difference is the conversion function called (watsonx_to_completion vs watsonx_to_chat_completion_chunk) and the return type. Consider unifying them with a generic approach or accepting a conversion closure to avoid the duplication.

src/clients/openai.rs
```diff
@@ -92,23 +125,418 @@ impl OpenAiClient {
...
+
+    async fn completions_via_router(
+        &self,
+        request: CompletionsRequest,
+        mut headers: HeaderMap,
+    ) -> Result<CompletionsResponse, Error> {
+        let router = self
+            .router_config
+            .as_ref()
+            .expect("router config must be present");
```
evaline-ju: These `.expect()` calls may panic and bring down the task if there's a logic error. It would be safer to return an actual Error

src/clients/openai.rs
```diff
@@ -92,23 +125,418 @@ impl OpenAiClient {
...
+
+            let url = router_client.endpoint(ROUTER_HANDLE_ENDPOINT);
+            let chat_completion = self
+                .handle_unary_with_body(router_client, url, http_pass_payload, headers)
+                .await?;
+            Ok(ChatCompletionsResponse::Unary(chat_completion))
+        }
+    }
+
+    // pub async fn tokenize(
```
evaline-ju: nit: Leftover commented-out code — should be removed

src/clients/openai.rs
```diff
@@ -92,23 +125,418 @@ impl OpenAiClient {
...
+            "x-reply-type",
+            router_config.reply_type.parse().map_err(|e| Error::Http {
+                code: StatusCode::BAD_REQUEST,
+                message: format!("failed to set x-reply-type header: {e}"),
+            })?,
+        );
+
+        // x-global-transaction-id: Unique request ID
+        let transaction_id = {
+            let tid = Uuid::new_v4().to_string();
```
evaline-ju: Without much context, this method always generates a new UUID for x-global-transaction-id without checking if one already exists in the incoming headers. The other two router methods (completions_via_router, tokenize_via_router) check for an existing header first and reuse it. This inconsistency could break request tracing/correlation. Should this match the pattern used in the other methods?

src/clients/openai.rs
```diff
@@ -92,23 +125,418 @@ impl OpenAiClient {
...
+                    if key != "detectors" && key != "_prompt_masks" {
+                        obj.insert(key.clone(), value.clone());
+                    }
+                }
+            }
+
+            let vllm_json_bytes = serde_json::to_vec(&vllm_request).map_err(|e| Error::Http {
+                code: StatusCode::INTERNAL_SERVER_ERROR,
+                message: format!("failed to serialise vLLM request: {e}"),
+            })?;
```
evaline-ju: serde_json::to_vec produces a Vec<u8>, and when that gets serialized into the
http_pass_payload JSON object, it becomes a JSON array of numbers: "payload": [123, 34, 109, ...]. Is this what the router sender expects? If it expects a string body or a nested JSON object, this would be a runtime bug. Could you confirm the expected encoding (raw bytes, base64 bytes, JSON obj)?

src/orchestrator.rs
```diff
@@ -145,11 +149,19 @@ async fn create_clients(config: &OrchestratorConfig) -> Result<ClientMap, Error>
         }
     }
 
+    let router_config = config.openai.as_ref().and_then(|o| o.router.clone());
```
evaline-ju: This ties detector routing to the OpenAI config's existence. If someone configures detectors without an `openai` block, the router will silently be disabled for detectors even if that wasn't the intent.

src/config.rs
```diff
@@ -151,13 +151,46 @@ pub struct GenerationConfig {
...
 /// OpenAI service configuration
 #[derive(Default, Clone, Debug, Deserialize)]
 pub struct OpenAiConfig {
     /// Generation service connection information
     pub service: ServiceConfig,
     /// Generation health service connection information
     pub health_service: Option<ServiceConfig>,
+    /// Optional router configuration. When present and `enabled: true`, the orchestrator
+    /// routes model calls through the router sender sidecar via Redis queues.
+    pub router: Option<RouterConfig>,
```
evaline-ju: This config is nested under OpenAiConfig, yet in orchestrator.rs it gets cloned and applied to every DetectorClient as well. If the router is intended to be shared across both model and detector calls, it might make more sense to place it at the OrchestratorConfig level rather than nesting it under openai and then pulling it out. As-is, if openai isn't configured but detectors are, router_config will silently be None

src/clients/openai.rs
```diff
@@ -356,8 +1136,10 @@ pub struct CompletionsRequest {
     #[serde(skip_serializing_if = "Option::is_none")]
     pub stream: Option<bool>,
     /// Model name.
+    #[serde(skip_serializing_if = "String::is_empty")]
     pub model: String,
     /// Prompt text.
+    #[serde(skip_serializing_if = "String::is_empty")]
```
evaline-ju: The #[serde(skip_serializing_if = "String::is_empty")] additions to model and prompt change serialization for all code paths, not just the router path. An empty model/prompt would now be silently omitted from the serialized output, which could mask bugs or produce confusing downstream errors. Was this intentional, or is it only needed for the router request construction?

src/clients/openai.rs
```diff
@@ -207,6 +635,358 @@ impl OpenAiClient {
...
+                                                let _ = tx.send(Ok(Some(completion))).await;
+                                            }
+                                            Err(serde_error) => {
+                                                let error = Error::Http {
+                                                    code: StatusCode::INTERNAL_SERVER_ERROR,
+                                                    message: format!(
+                                                        "router stream deserialization error: {serde_error}"
+                                                    ),
+                                                };
+                                                let _ = tx.send(Err(error.into())).await;
```
evaline-ju: On a deserialization error, this sends Err to the channel but does not break out of the loop. Compare with the chat handler at line 843-844 which correctly does break after the send.

Without a break, the completions stream will:
1. Keep processing events after a corrupt chunk
2. Eventually exit the loop and send Ok(None) at line 758 — making it look like the stream completed successfully

src/clients/openai.rs
```diff
@@ -207,6 +635,358 @@ impl OpenAiClient {
...
+                                        }
+                                    }
+                                }
+                            }
+                            Err(error) => {
+                                let error = Error::Http {
+                                    code: StatusCode::INTERNAL_SERVER_ERROR,
+                                    message: error.to_string(),
+                                };
+                                let _ = tx.send(Err(error.into())).await;
```
evaline-ju: When the SSE event stream itself errors, this sends Err but does not break. The loop continues trying to read from a potentially broken stream. The chat handler at line 855-856 correctly breaks here. Should add break here

src/clients/openai.rs
```diff
@@ -207,6 +635,358 @@ impl OpenAiClient {
...
+                                break;
+                            }
+                        }
+                    }
+                    // Stream ended — signal completion
+                    let _ = tx.send(Ok(None)).await;
```
evaline-ju: Both stream handlers unconditionally send Ok(None) after the while loop exits (line 758 and line 861), but every normal termination path ([DONE], "err" event) already sends its own signal and breaks. This means every stream sends a double termination. Not a crash — the second send will likely fail silently since the receiver may have dropped — but combined with the missing breaks above, it means the completions handler can send Err followed by Ok(None), which a consumer could misinterpret as a partial success.

src/clients/openai.rs
```diff
@@ -207,6 +635,358 @@ impl OpenAiClient {
...
+                                                &event.data,
+                                            ) {
+                                                Ok(envelope) => {
+                                                    match envelope
+                                                        .get("content")
+                                                        .and_then(|v| v.as_str())
+                                                    {
+                                                        Some(s) => s.to_string(),
+                                                        None => {
+                                                            if event.data == "[DONE]" {
```
evaline-ju: consideration: The [DONE] check only runs when the envelope has no content field (line 707) or isn't valid JSON (line 717). If the router ever wraps [DONE] in an envelope like {"content": "[DONE]"}, it would fall through to the Watsonx deserialization and produce a spurious error. The chat handler (line 802) avoids this by checking event.data == "[DONE]" before parsing the envelope. Consider using the same approach here for robustness.

evaline-ju (commented):
Thanks for the contribution! Due to the size of the PR I’ve leveraged some AI tools for review

In addition to inline comments, there’s a lot of new logic added here - while showing the current tests pass is good, it would be good practice to add new tests for the new additions. Some suggestions:
- Unit tests for watsonx_to_completion / watsonx_to_chat_completion_chunk — these are pure functions and trivially testable (various stop_reason values, empty results, missing model_id)
- Unit tests for RouterConfig deserialization — defaults, enabled/disabled, partial config
- Unit tests for http_pass_payload construction — verifying the correct structure, path, and headers
- Tests for the stream handlers — "png" keepalive skipped, "err" events produce errors, "[DONE]" terminates cleanly, malformed data is handled gracefully

src/clients/openai.rs
gopikaa-rajan: Added a common method: `build_router_headers` for common header construction logic

src/clients/openai.rs
gopikaa-rajan: This is addressed...replaced all .expect() calls with an error

src/clients/openai.rs
gopikaa-rajan: This is removed

src/clients/openai.rs
gopikaa-rajan: Created Generic Helper Method:
Added handle_router_stream<T>()  - A generic method that accepts a conversion closure to handle both completion types

Accepts a converter function: impl Fn(WatsonxTextGenResponse, &str) -> T + Send + 'static
Refactored Both Streaming Methods:

handle_router_completions_stream() - calls generic helper with watsonx_to_completion converter
handle_router_chat_completions_stream() - calls generic helper with watsonx_to_chat_completion_chunk converter

src/clients/openai.rs
gopikaa-rajan: The `build_router_headers()` method correctly checks for an existing x-global-transaction-id header first now..

src/clients/openai.rs
gopikaa-rajan: The expected encoding is base64... have handled this by converting to base64

src/orchestrator.rs
gopikaa-rajan: Kept the router config at orchestrator level rather than within openai

src/clients/openai.rs
gopikaa-rajan: This is removed now..

src/clients/openai.rs
gopikaa-rajan: done

src/clients/openai.rs
gopikaa-rajan: done

src/clients/openai.rs
gopikaa-rajan: done

src/clients/openai.rs
gopikaa-rajan: Removed the unconditional Ok(None) send after the loop..

evaline-ju (commented):
The tests placed in the `tests/` dir are expected to be integration tests that spin up mock servers and test against them. 

The added tests appear to mostly be unit tests, which in Rust belong in the same file as the source, but many of them do not appear to test anything from the source code. 

For example stream_handler_tests appear to be creating a mock event with a create_event function in the tests and just checking the mock structure. Similarly the http_pass_payload_tests appear to just be creating a json structure and checking attributes.

Please make sure added tests are actually testing parts of the code so that they can be depended on to check the code.

Sanj180706 (approved):
LGTM

PR #536, opened 2026-06-25
Router Integration with Selective Detector Support + Streaming Empty Output Fix

gopikaa-rajan:
This PR does below:

1. ConfigMap enhancement for granular control over which detectors use the router service
2. Resolution of chunker failures when models return empty text in streaming mode


Sanj180706 (approved):
LGTM

